PR #141, opened 2023-03-20
Updating Jinja descriptions

thomasstvr:
All tests passed except for S3 (because of log in credentials?)


ChrisWRWX:
Can you add some tests to this before merging?

PR #208, opened 2023-07-03
Updated schema to correct format

thomasstvr:
When auto-filling in recipe, default was returned without ':'. Updated schema, and added 'type: string' to default.


thomasstvr:
Need to update schema to show all types default can take on

PR #221, opened 2023-07-19
Added meaningful error message to data.py

thomasstvr:
Added the bit of code below to check for status code 400 and respond with meaningful error message:
```  
if response.status_code == 200:
    results = response.json()
elif response.status_code == 400:
    raise RuntimeError('Not able to validate access, check wrangle to ensure accessibility')
```


ChrisWRWX:
Does this give a 400? 401 or 403 are the correct responses for not authorized.

If this is giving a 400 then I should fix the back-end to give a better error status.

This would probably be a better syntax in case of not specifically 400 errors.

```py
if response.ok:
  # do stuff
elif response.status_code in [401, 403]:
  raise unauthorized
else:
  raise generic error
```

thomasstvr:
Yes, it gives a 400

ChrisWRWX (approved):

PR #224, opened 2023-07-20
Added model_id to wrong type error message

thomasstvr:
Added model_id to error message for classify, standardize and extract


ChrisWRWX:
Please add two corresponding tests for extract and classify

thomasstvr:
Added tests and fixed model_id to list in extract.py

ChrisWRWX (approved):

Issue #57, opened 2022-08-09
Remove Words Capitalization


cerdamario13:
Look at Remove_words and see if it can
- preserve case
- inputs as raw (no case conversion)

cerdamario13:
Also look into what happens when a list to remove is empty

cerdamario13:
Fixed the error out of range
The issue was not tokenizing correctly if using `tokenize_to_remove`
`[tokenize(to_remove[x]) for x in range(len(to_remove))]`

cerdamario13:
@ebhills 

Fixed out of range index (see above comment) error.
Added `ignore_case` as a parameter. If set to False, the input and to_remove will be taken as Raw tokens.

```yaml
- remove_words:
      input: Column1
      to_remove:
        - materials
        - Case Colours
      output: Item
      tokenize_to_remove: True
      ignore_case: True
```

ebhills:
thx. On your ignore_case argument, would like to clarify and maybe suggest a different name. Does True mean that it matches on 'as is' case in the inputs?  If so, you might want to name the argument something like case_sensitive.  Because I assume that your False logic will match tokend regardless of case (i.e. case insensitive).

cerdamario13:
no, True means that it takes values and convert all to same case.

I will change the naming to 'case_sensitive'
- True: the Wrangle will convert all values to same case 
- False: the Wrangle will take data 'as-is'

Let me know if that works

cerdamario13:
Updated to case_sensitive instead of ignore_case
The capitalization is also removed -> The output was set to title case instead of leaving it as it was

example with case_sensitive and tokenize set to true


Test   Input | Extract Voltage | Extract Shape | Extract Colours | Output
-- | -- | -- | -- | --
my list of thInGs: ROunD,   squarE | ROunD,   Square |  | | my   list of thInGs: squarE
blue, GrEEn t-shIRt |   |   | Blue,   GrEEn | blue   t-shIRt
20 V poWer sAW | 20   V |   |   | poWer sAW

PR #284, opened 2023-09-13
Remove capitalization remove words

cerdamario13:
Renaming ingnore_case to case_sensitive and removing title casing of the outputs. Test have been updated.

Used regex sub instead of lists and tokenize to remove words. This is the same as Remove Words in Excel. Now the case is kept as is. This is related to issue #57 

e.g.

Test   Input | Extract Voltage | Extract Shape | Extract Colours | Output
-- | -- | -- | -- | --
my list of thInGs: ROunD,   squarE | ROunD,   Square |  | | my   list of thInGs: squarE
blue, GrEEn t-shIRt |   |   | Blue,   GrEEn | blue   t-shIRt
20 V poWer sAW | 20   V |   |   | poWer sAW


ChrisWRWX:
Is there a reason to rename the variable? That makes it a breaking change.

It's understandable either way so I'd lean towards leaving it as the old name unless you have a good reason otherwise.

cerdamario13:
Eric wanted the name change. I would like to keep it as is personally

ChrisWRWX:
Leave it as the old one then. No point breaking things for the sake of it.

As it stands you've also flipped the default. ignore_case was true, now case_sensitive is true. Make the default that case is insensitive.

ChrisWRWX (approved):

PR #295, opened 2023-09-26
added google drive connector

cerdamario13:
Google drive connector

Here is some information about setting a service account
 - More information regarding auth and keys
 - https://developers.google.com/identity/protocols/oauth2/service-account
 - be sure to also share the folder with the service account email address as Editor

Here is how to set the env variables in python. I can share the data in slack or you can make an account
```python
projectID = os.getenv("GOOGLe_PROJECT_ID", "...")
privateKeyID = os.getenv("GOOGLE_PRIVATE_KEY_ID", "...")
privateKey = os.getenv("GOOGLE_PRIVATE_KEY", "...")
clientEmail = os.getenv("GOOGLE_CLIENT_EMAIL", "...")
clientID = os.getenv("GOOGLE_CLIENT_ID", "...")

read_file_id = "file_ID" # data.xlsx file

df_data = read(
        file_id=read_file_id,
        project_id=projectID,
        private_key_id=privateKeyID,
        private_key=privateKey,
        client_email=clientEmail,
        client_id=clientID
        )


write(
    df=df_data,
    folder_id='folder_ID',
    file_name='write_file.xlsx',
    project_id=projectID,
    private_key_id=privateKeyID,
    private_key=privateKey,
    client_email=clientEmail,
    client_id=clientID,
    to_sheets=True
)
```

for reading, a file id must be provided. Here is more [information ](https://developers.google.com/sheets/api/guides/concepts)about file IDs

For writing, a folder ID must be provided.

We can set up a shared folder that can contain a testing folder for reading and writing

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -41,6 +41,11 @@ jobs:
           CKAN_API_KEY: ${{ secrets.CKAN_API_KEY }}
           HUGGINGFACE_TOKEN: ${{ secrets.HUGGINGFACE_TOKEN }}
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          GOOGLE_PROJECT_ID: "drive-connector-361914"
+          GOOGLE_CLIENT_ID: "112123640790185424876"
+          GOOGLE_CLIENT_EMAIL: "drive-connector@drive-connector-361914.iam.gserviceaccount.com"
+          GOOGLE_PRIVATE_KEY_ID: "4cd1686b2da8ab0c9ea30dc608f070b7ade78d28"
+          GOOGLE_PRIVATE_KEY: ${{ secrets.GOOGLE_PRIVATE_KEY }}
 
   test-pip-install:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


cerdamario13:
Here are the folders that can be deleted from the google imports:
 - google_api_core-2.12.0.dist-info
 - google_api_python_client-2.101.0.dist-info
 - google_auth_httplib2-0.1.1.dist-info
 - google_auth-2.23.0.dist-info
 

Folders that can be deleted from **google** folder:
  - _upb
  - api
  - cloud
  - gapic
  - logging
  - protpbuf
  - rpc
  - type
  

In **googleapiclient**, navigate to discovery_cache folder -> documents. Here delete everything except Drive files:
  - drive.v2
  - drive.v3
  - dtiveactivity.v2
  - drivelabels.v2
  - drivelabels.v2beta

This should reduce the package size to ~5 MB.

ChrisWRWX:
Is there an api endpoint that will translate the file path + name to an ID? It would be significantly better to be able to just pass those rather that having to figure out an ID, if possible.

ChrisWRWX:
- [ ] Needs the schema definitions adding.
- [ ] Needs tests adding.

ChrisWRWX:
> Here is how to set the env variables in python. I can share the data in slack or you can make an account
> 
> ```python
> projectID = os.getenv("GOOGLE_PROJECT_ID", "...")
> privateKeyID = os.getenv("GOOGLE_PRIVATE_KEY_ID", "...")
> privateKey = os.getenv("GOOGLE_PRIVATE_KEY", "...")
> clientEmail = os.getenv("GOOGLE_CLIENT_EMAIL", "...")
> clientID = os.getenv("GOOGLE_CLIENT_ID", "...")
> ```

Not sure I understand what you're getting at here. The connector appears to have the various credentials defined as parameters, as it should. Is there some reason you have to use these env variables specifically?

```yml
read:
  - google_drive:
      privateKey: blablabla
      clientID: blablabla
      ...
```

cerdamario13:
> Is there an api endpoint that will translate the file path + name to an ID? It would be significantly better to be able to just pass those rather that having to figure out an ID, if possible.

Added so that we can use the sharable link or the ID of the file/folder. I think this is a common way to reference a file in drive. 
image

I will look into using the file path

cerdamario13:
I added tests. Now to add the secrets so that it passes the github actions
The tests are using real files and a folder that WrWx members have access. If we encounter issues, we can always mock the tests.
The folder will fill up with files, I have a script that can clear the folder that I need to set to run periodically. I had issues with trying to clear the folder  it after the tests.

ChrisWRWX:
Great work! Does this work for write too?

cerdamario13:
> Great work! Does this work for write too?

Yes, It works for both read and write.
Also when writing, if there is already file with the name you are trying to write it overwrites that file. I could set this a a toggle if we want

ChrisWRWX:
Cool - yeah, meant read. Your example was write.

Let's discuss Friday and sort the test cases and then get it merged.

ChrisWRWX:
Closing this. May come back to in future but currently low priority.

Issue #294, opened 2023-09-25
A more succinct option for creating columns


cerdamario13:
This will look something like this.

```yaml
    wrangles:
      - create.column:
          output:
            - new1: val1
            - new2: val2
```

This is what I have added to create.column function and so fart it passed all of the tests

```python
        try: 
            value = [list(values.values())[0] for values in output]
            output = [list(keys.keys())[0] for keys in output]
        except:
            pass
```

Here is in the function
```python
def column(df: _pd.DataFrame, output: _Union[str, list], value = None) -> _pd.DataFrame:
    """
    type: object
    description: Create column(s) with a user defined value. Defaults to None (empty).
    additionalProperties: false
    required:
      - output
    properties:
      output:
        type:
          - string
          - array
          - object
        description: Name or list of names of new columns
      value:
        type: string
        description: (Optional) Value to add in the new column(s). If omitted, defaults to None.
    """    
    
    # check if the input is a list and check if elements are a dictionary
    if isinstance(output, list):
        try: 
            value = [list(values.values())[0] for values in output]
            output = [list(keys.keys())[0] for keys in output]
        except:
            pass
    
    # Get list of existing columns
    existing_column = df.columns
    
    # Get number of rows in df
    rows = len(df)
    # Get number of columns created
    cols_created = len(output)
    # If a string provided, convert to list
    if isinstance(output, str):
      if output in existing_column:
        raise ValueError(f'"{output}" column already exists in dataFrame.')
      output = [output]
    
    # Allow for user to enter either a list and/or a string in output and value and not error
    if isinstance(value, list) and cols_created == 1:
        value = [value[0] for _ in range(cols_created)]
    elif isinstance(value, str):
        value = [value for _ in range(cols_created)]
    elif value == None:
        value = ['' for _ in range(cols_created)]
        
    # Check if the list of outputs exist in dataFrame
    check_list = [x for x in output if x in existing_column]
    if len(check_list) > 0:
      raise ValueError(f'{check_list} column(s) already exists in the dataFrame') 
    
    for output_column, values_list in zip(output, value):
        # Data to generate
        data = _pd.DataFrame(_generate_cell_values(values_list, rows), columns=[output_column])
        data.set_index(df.index, inplace=True)  # use the same index as original to match rows
        # Merging existing dataframe with values created
        df = _pd.concat([df, data], axis=1)

    return df
```

ChrisWRWX:
I like this solution, but I think the code needs a bit more iteration.

[...]

PR #301, opened 2023-10-04
A more succinct option for creating columns #294

cerdamario13:
#294 

Dictionaries can now used in the output of `create.column` for a more succinct recipe.

```yaml
- create.column:
     output:
        col2: val2
        col3: val3
```

A value can be set that applies to any not defined directly within the output

```yaml
- create.column:
     output:
        - col2: val2
        - col3
        - col4
     value: This is a test
```

### DIFF
--- a/tests/connectors/test_test.py
+++ b/tests/connectors/test_test.py
@@ -110,7 +110,49 @@ def test_random_list():
       - test:
           rows: 3
           values:
-            random_list: random<["true", "false"]>
+            random_list: <random(["true", "false"])>
     """
     df = wrangles.recipe.run(recipe)
     assert df.iloc[0][0] == 'true' or df.iloc[0][0] == 'false'
+    
+def test_random_list_objects():
+    """
+    Test using an object in random list
+    """
+    recipe = """
+    read:
+      - test:
+            rows: 3
+            values:
+                random_list: <random([{"a":1}, {"b":2}])>
+    """
+    df = wrangles.recipe.run(recipe)
+    assert df.iloc[0][0] == {'a': 1} or df.iloc[0][0] == {'b': 2}
+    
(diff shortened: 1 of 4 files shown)


ChrisWRWX:
Historically there's been an option to pass a list and it randomly chooses from the values list. I think it's a little confusing though, so as part of this can you edit it so that a list just becomes a list within the created cell?

Can we make the random choice behave like the others? `<random(["one","two","three"])>`

That's probably within the test connector.

cerdamario13:
Something like this?

```python
df = wrangles.recipe.run(
    recipe = """
    read:
      - test:
          rows: 5
          values:
            Order_List:
              - apple
              - orange
              - banana
            Random_list: random<["one", "two", "three"]>
    """
)
```

| Order_List                    | Random_list   |
|:------------------------------|:--------------|
| ['apple', 'orange', 'banana'] | three         |
| ['apple', 'orange', 'banana'] | one           |
| ['apple', 'orange', 'banana'] | three         |
| ['apple', 'orange', 'banana'] | two           |
| ['apple', 'orange', 'banana'] | two           |

ChrisWRWX:
Yeah

cerdamario13:
I have added the `random list ` and `order list` to the test connector and updated tests
Forgot to mention that the test connector works with the new create columns style

```yaml
- create.column
    output: 
        - col2: <int>
        - col3: regular string
```

ChrisWRWX:
A correction on the random syntax.

Should be `<random(["one","two","three"])>` to match the style of the existing ones.

https://wrangles.io/python/connectors/test#special-inputs

Any JSON array should work within the brackets so `([1,2,3])` or `([true, false])` or `([{}, [], ...])`

ChrisWRWX (approved):

PR #316, opened 2023-10-26
Split multiple dictionaries

thomasstvr:


ChrisWRWX:
- [ ] Add a test case that tests for duplicated keys in the input data.
- [ ] Simplify code to assume a list and just loop over the list. Ensure that it behaves correctly if a key is duplicated, and that the first/last is kept.

ChrisWRWX (changes requested):
- [x] Add a test case that tests for duplicated keys in the input data.
- [x] Simplify code to assume a list and just loop over the list. Ensure that it behaves correctly if a key is duplicated, and that the first/last is kept.

ChrisWRWX (approved):

PR #317, opened 2023-10-27
Unit conversion schema and tests

cerdamario13:
Added conversion units to `extract.attributes` schema
Added conversion tests
`desired unit` works both with singular or plural. Used singular for schema.


ChrisWRWX (changes requested):
Move large attribute_type / unit conversion map to be a part of the base schema and reference it instead, then we can re-use this for standardize/convert that we'll add in future.

cerdamario13:
> Move large attribute_type / unit conversion map to be a part of the base schema and reference it instead, then we can re-use this for standardize/convert that we'll add in future.

Done

ChrisWRWX (approved):

PR #323, opened 2023-11-01
Cosine similarity

thomasstvr:


thomasstvr:
Added euclidean distance (with tests) and streamlined code with numpy arrays.

ChrisWRWX:
- [ ] Add a test that ensures it works with vectors as numpy arrays in addition to lists.
- [ ] Use input with a list for both inputs rather than input1, input2. That way the user can use the dynamic column features (wildcards, regex) if they don't know the exact name. Validate that the user has provided exactly two columns, you can specify the number of items in JSON schema too.
- [ ] Set a default of cosine for type.
- [ ] Add an enum for type in the schema.
- [ ] Rename type to method. Type is sometimes problematic due to being a keyword.
- [ ] Make cosine the normal cosine calculation. Add another named 'adjusted cosine', and explain that it adjusts the typical cosine range to 0-1.
- [ ] Rewrite loops as list comprehensions. They're typically faster to execute than regular loops.
- [ ] Add underscore for numpy import.

thomasstvr:
I love the check boxes, I'll take care of this tomorrow.

ChrisWRWX (approved):

PR #355, opened 2023-12-13
Extract first element (for stock wrangles)

thomasstvr:


ChrisWRWX:
To be replace with decorators

PR #368, opened 2023-12-19
Added use_spellcheck to extract.custom

thomasstvr:


thomasstvr:
It seems that all of the translate tests have failed because of yesterday's fix.

ChrisWRWX (approved):

PR #417, opened 2024-02-05
Extract first element

thomasstvr:


thomasstvr:
See pull request #355

PR #427, opened 2024-02-12
Jinja characters

thomasstvr:


ChrisWRWX:
- [ ] Simplify code to regex replace non alphanumeric or underscore with underscore, rather than a fixed list of disallowed characters.

ChrisWRWX:
- [ ] Use native re instead of regex package

ChrisWRWX (approved):

PR #513, opened 2024-06-11
Flatten Table wrangle and test cases

v2ngu:
Created a Flatten Table wrangle with test cases

```yaml
wrangles:
  - flatten_table:
       similar: 2 # Optional param, matching accuracy of subheader
       case_sensitive: True # Optional subheadings case matching
```


ChrisWRWX:
- [ ] Allow the user to provide a column to override and indicate explicitly which are the header rows. Should be any 'truthy' value i.e. True/False but also any string vs empty string. If not provided, uses the current behaviour of trying to identify them.
- [ ] Should follow the format <verb>.<noun> - not a fan of flatten. Perhaps align.columns?
- [ ] Add an optional output parameter that allows the user to set the final output columns. If not specified, will include all found as now. If specified, will result in those columns in that order, and anything not matching will be discarded.
- [ ] It looks like the existing code will always result in lower case headers if you specifiy case insensitive. Can that be changed such that it just takes the case from the first instance?

PR #515, opened 2024-06-14
Extract Raw and selecting bracket param

v2ngu:
Two new parameters for extract brackets that work in tandem. 

- Find (list or str) of brackets to look for
- Extract raw returns what was found raw

```
wrangles:
   - extract.brackets:
       input: Input A
       output: Output A
       find: 
            - curly
            - round
       extract_raw: true
```


cerdamario13 (changes requested):
- [x] rename `extract_raw` param to `include_brackets`. This gives more information as to what it does.
- [ ] find param. Maybe rename to `bracket_type`? We can discuss with Thomas and Chris and see what they think

ChrisWRWX (approved):

cerdamario13 (approved):

PR #519, opened 2024-06-14
Handling where clause error

v2ngu:
Handles the error for the where clause only if the where returns an empty dataframe.


ChrisWRWX:
Closing this. On reflection, and following some discussions, it's not possible to solve this universally. By short circuiting it, the wrangles will never be executed and this will likely cause a following wrangle to fail in the common scenario that the output becomes the input of the next wrangle. If the code is simply skipped, the wrangle will not make the output column required.

This will instead be fixed by testing the wrangles to identify the ones currently failing and fixing them individually.

PR #532, opened 2024-07-03
364 overlap wrangle

cerdamario13:
Using pythons diff lib to find the similarity match of two columns

```yaml
wrangles:
    - find_match:
        input_a: col1
        input_b: col2
        output: TEST
```

| col1     | col2       | TEST            |
|:---------|:-----------|:----------------|
| mario    | supermario | *****mario      |
| luigi    | superluigi | *****luigi      |
| peach    | superpeach | *****peach      |
| yoshi    | superyoshi | *****yoshi      |
| equal    | equal      |EXACT_MATCH |
|          | no match   | A EMPTY       |
| no match |            | B EMPTY     |
|          |            | BOTH EMPTY  |

PR #533, opened 2024-07-03
463 compare wrangle

cerdamario13:
New module `compare`
New Wrangle `compare.text`: Compare two strings and return the intersection or difference, or use overlap to find the matching characters between the two strings.

**Difference**

```yaml
    wrangles:
    - compare.text:
        input:
          - col1
          - col2
        output: output
        method: difference
        char: ' ' # Optional character to split text
```

| col1   | col2        | output   |
|:-------|:------------|:---------|
| Mario  | Super Mario | Super    |
| Luigi  | Super Luigi | Super    |

**Intersection**

```yaml
    wrangles:
    - compare.text:
        input:
          - col1
          - col2
        output: output
        method: intersection
       char: '  ' # Optional character to split text
```

| col1   | col2        | output   |
|:-------|:------------|:---------|
| Mario  | Super Mario | Mario    |
| Luigi  | Super Luigi | Luigi    |

**Overlap**

```yaml
    wrangles:
    - compare.text:
        input:
          - col1
          - col2
        output: output
        method: overlap

        non_match_char: '@' # (Optional) Character to use for non-matching characters
        include_ratio: True # (Optional) Include the ratio of matching characters
        decimal_places: 2 # (Optional) Number of decimal places to round the ratio to
        exact_match: 'They are the same' # "(Optional) Value to use for exact matches. Default is empty string
        empty_a: 'Empty A' # (Optional) Value to use for empty input a. Default is empty string
        empty_b: 'Empty B' # (Optional) Value to use for empty input b. Default is empty string
        all_empty: 'Both Empty' # (Optional) Value to use for both inputs. Default is empty string
```

| col1   | col2       | output     |
|:-------|:-----------|:-----------|
| Mario  | SuperMario | *****Mario |
| Luigi  | SuperLuigi | *****Luigi |
| Toad   |            |            |


ChrisWRWX:
- [ ] Wrangles should use the syntax <verb>.<noun> to reflect the actual underlying code structure, please nest this properly rather than using an underscore.
- [ ] Use input rather than input_a, input_b. That way you can use all the features for input - wildcards, optional columns etc. Validate that 2 columns are provided if the method requires exactly 2 inputs.
- [ ] What's the difference between this and #532? They all seem to be doing roughly the same thing. Merge them and replace 'type' with a 'method' parameter than can apply the different specific methods.

ChrisWRWX:
- [ ] Tidy unnecessary import
- [ ] Moved schema fully to docstring
- [ ]

ChrisWRWX (approved):

ChrisWRWX (approved):

PR #542, opened 2024-07-23
Standardize attributes

cerdamario13:
Standardize and remove attributes

- the new standardize and remove wrangles are in `format`, should there be a `standardize` folder?
- moved the extract conversion tests to test_extract in a class.
- A lot of the object tests had to be updated since there is a new key `standard`


cerdamario13:
Work still in progress, will reopen when ready

PR #546, opened 2024-08-05
Standardize attributes

cerdamario13:
Removed breaking change
~~**BREAKING CHANGE**~~

~~Updated standardize module to include attributes and custom.~~
~~python wrangle `wrangles.standardize` no longer exists, use `wrangles.standardize.custom` instead~~

**New extract attributes optional parameters**
params:

- responseContent: standard, this returns a standardize version of the attribute
- sigFigs: apply the significant figures to the attribute only if converting units. This is due to not interfering with span (trick -> set attribute_type and desired_unit to the same attribute type to apply significant figures)

```yaml
        wrangles:
            - extract.attributes:
                input: Tools
                output: Attributes
                responseContent: standard
                attribute_type: length
                sigFigs: 3 # optional
```

**New wrangle**
standardize.attributes: standardizes attributes in text. 
params:

- attribute_type if is not defined, then all attributes are standardized,
- desiredUnit: convert the attribute type to a specific unit
- sigFigs: set the number of significant figures only if converting units. (trick -> set attribute_type and desired_unit to the same attribute type to apply significant figures)

```yaml
        wrangles:
            - standardize.attributes:
                input: Tools
                output: Attributes
                attribute_type: weight # optional
                desired_unit: meter # opitonal
                sigFigs: 3 # optional
```


cerdamario13:
- [ ] Remove the breaking change -> `wrangles.standardize` should still work.

PR #568, opened 2024-09-24
Lookup schema/Tests

thomasstvr:
I feel like I am missing something here....


ChrisWRWX (dismissed):

ChrisWRWX:
- [ ] Add validation for write lookup. If model_id not variant or name. If not model_id must have name, variant should default to key. May already exist.

ChrisWRWX (approved):

PR #640, opened 2025-01-07
Added better error handling to round function

kavu16:
Added better error handling to fix an issue that was occurring with the rounding recipe.  

The recipe used the default behavior of the pandas `round` function, returning the original column if any of the row types were incompatible.  

Added the `to_numeric` function, which allows coercion to a float and returns `nan` if the datatype is incompatible.  Then the pandas `round` function is applied as normal and any remaining `nan` is replaced with an empty string to play nice with the front end

### DIFF
--- a/tests/recipes/wrangles/test_pandas.py
+++ b/tests/recipes/wrangles/test_pandas.py
@@ -438,11 +438,62 @@ class TestRound:
                 decimals: 1
             """,
             dataframe=pd.DataFrame({
-                'col': [3.13, "Something else", 1.16, None]
+                'col': [3.13, "Something else", 1.16, None, "2.5555"]
             })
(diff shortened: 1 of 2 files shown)


ChrisWRWX:
I feel like I've seen some issues with round and numpy floats. Please will you add a few more tests with different numpy floats? Also perhaps add a couple with numbers as strings '3.14159'. Make sure they also give the correct values back.

kavu16:
Added different numpy float aliases to test to ensure proper behavior.  
Current function maps everything back to a standard python `float`.  Might be better to return the original datatype, but users have run into strange issues when getting numpy `float`s returned

ChrisWRWX (approved):

PR #644, opened 2025-01-07
Sentence case fix

kavu16:
Changed functionality of `convert.case` function to capitalize each sentence in a string rather than just the first.  

Used `split` function and then mapped `lstrip()` and `capitalize()` to each sentence in the list, stripping whitespace at the end.

This should fix the issue where only the first sentence was capitalized.  Added benefit for cleaning up weirdly formatted paragraphs, though this may be something to discuss later.


ChrisWRWX:
This shouldn't have strip included - it shouldn't alter the user's spacing. We can't assume the spacing isn't intentional. The strings should still be equal, only the case should change after the wrangle.

There are also more edge cases to consider...

- My 13.5mm bolt.
- will this be capitalized correctly? i'm not so sure! probably not... but maybe

ChrisWRWX (approved):

PR #645, opened 2025-01-08
Extract regex capture groups

### LINKED ISSUE #502: extract.regex - How do you specify capture groups?
Need ability to specify capture groups like \1 \2 etc.  But the output filed is just the column name. Maybe add a param called return?

thomasstvr:
This will close #502 with some small additions/omissions...

- First_element parameter was left out in anticipation of decorators.
- A bug was fixed where only capture groups were output for any pattern with a capture group. ie "55" output for "(\d+).*" instead of "55g".
- Unsure about the parameter name capture_group. Might be too technical, but then again this wrangle is more advanced than most.


thomasstvr:
- [ ] Add tests for outputting multiple capture groups, make changes if needed.
- [ ] Other stuff I forget

ChrisWRWX (changes requested):
This is still pending the changes we discussed for the syntax of the capture groups, right?

ChrisWRWX (approved):

PR #647, opened 2025-01-09
Split wildcard bug

### LINKED ISSUE #608: Split.list Wildcard Breaks When Passed as a List
When using a wildcard and passing the output column as a list there is a value error that is thrown:

"Exception has occurred: ValueError
split.list - Columns must be same length as key"

```yaml
  - split.list:
      input: column
      output: 
        - Col*
```

thomasstvr:
This PR will close #608


thomasstvr:
- [ ] Look into using the wildcard_expansion function instead

ChrisWRWX (approved):

PR #648, opened 2025-01-10
Hash recipe function + testing

kavu16:
Created recipe to hash inputs using the `hashlib` library.  

Input is converted to string, encoded to byte code representation using utf-8 standard, hashed, and then output as a hexdigest.

Currently supporting `md5`, `sha1`, `sha256`, and `sha512`, raising an exception if another hash method is requested.

### DIFF
--- a/wrangles/recipe_wrangles/create.py
+++ b/wrangles/recipe_wrangles/create.py
@@ -14,7 +14,7 @@ from jinja2 import (
 )
 from ..connectors.test import _generate_cell_values
 from .. import openai as _openai
-from hashlib import sha1 as _sha1, sha256 as _sha256, sha512 as _sha512, md5 as _md5
+import hashlib as _hashlib
 
 def bins(df: _pd.DataFrame, input: _Union[str, list], output: _Union[str, list], bins: _Union[int, list], labels: _Union[str, list] = None, **kwargs) -> _pd.DataFrame:
     """
@@ -423,25 +423,28 @@ def hash(df: _pd.DataFrame, input: _Union[str, list], output: _Union[str, list],
     additionalProperties: false
     required:
       - input
-      - output
     properties:
       input:
         type:
+          - string
           - array
         description: Name of input column
       output:
         type:
+          - string
           - array
         description: Name of new column
       method:
         type: string
-        description: The method to use to hash the input
+        description: The method to use to hash the input (Default: md5)
         enum:
           - md5
           - sha1
           - sha256
           - sha512
     """
+    if output is None: output = input
+
     if not isinstance(input, list): input = [input]
     if not isinstance(output, list): output = [output]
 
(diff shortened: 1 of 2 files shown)


wrangles/recipe_wrangles/create.py
```diff
@@ -414,3 +415,49 @@ def uuid(df: _pd.DataFrame, output: _Union[str, list]) -> _pd.DataFrame:
...
+    type: object
+    description: Create a hash of a column
+    additionalProperties: false
+    required:
+      - input
+      - output
+    properties:
+      input:
+        type:
+          - array
```
ChrisWRWX, line 430: Should allow string or array

wrangles/recipe_wrangles/create.py
```diff
@@ -414,3 +415,49 @@ def uuid(df: _pd.DataFrame, output: _Union[str, list]) -> _pd.DataFrame:
...
+        description: Name of input column
+      output:
+        type:
+          - array
```
ChrisWRWX, line 434: Should allow string or array

wrangles/recipe_wrangles/create.py
```diff
@@ -414,3 +415,49 @@ def uuid(df: _pd.DataFrame, output: _Union[str, list]) -> _pd.DataFrame:
...
     return df
+
+def hash(df: _pd.DataFrame, input: _Union[str, list], output: _Union[str, list], method: str = 'md5') -> _pd.DataFrame:
+    """
+    type: object
+    description: Create a hash of a column
+    additionalProperties: false
+    required:
+      - input
+      - output
```
ChrisWRWX, line 426: Output shouldn't be required.

wrangles/recipe_wrangles/create.py
```diff
@@ -414,3 +415,49 @@ def uuid(df: _pd.DataFrame, output: _Union[str, list]) -> _pd.DataFrame:
...
+        type: string
+        description: The method to use to hash the input
+        enum:
+          - md5
+          - sha1
+          - sha256
+          - sha512
+    """
+    if not isinstance(input, list): input = [input]
+    if not isinstance(output, list): output = [output]
```
ChrisWRWX, line 446: Standard behaviour for a wrangle is that if output is omitted, it overwrites the input.

Add `if output is None: output = input` and set default for output = None

wrangles/recipe_wrangles/create.py
```diff
@@ -414,3 +415,49 @@ def uuid(df: _pd.DataFrame, output: _Union[str, list]) -> _pd.DataFrame:
...
+
+    if len(input) != len(output):
+        raise ValueError('The lists for input and output must be the same length.')
+
+    for in_col, out_col in zip(input, output):
+        if method == 'md5':
+            df[out_col] = df[in_col].apply(lambda x: _md5(str(x).encode('utf-8')).hexdigest())
```
ChrisWRWX, line 453: This could be simplified using getattr. `hash_fn = getattr(hashlib, method)`

Won't need all the if, elif, elif... then.

You'll need to add the validation before that method is one of the allowed values, but that can be done prior once rather than being incorporated into the loop.

wrangles/recipe_wrangles/create.py
```diff
@@ -414,3 +415,49 @@ def uuid(df: _pd.DataFrame, output: _Union[str, list]) -> _pd.DataFrame:
...
+        type:
+          - array
+        description: Name of input column
+      output:
+        type:
+          - array
+        description: Name of new column
+      method:
+        type: string
+        description: The method to use to hash the input
```
ChrisWRWX, line 438: State the default here.

ChrisWRWX (changes requested):

ChrisWRWX (approved):

PR #688, opened 2025-03-21
Replace lists

### LINKED ISSUE #639: Replace Wrangle Turns Lists Into Strings
When running the replace wrangle on a list of strings, the list is converted to a string.

```
  - replace:
      input:
        - Features
      find: "N/A"
      replace: ""
```

This occurs without any errors making it difficult to find and debug. Possibly raise and error when running on anything other than a string or have a silent failure that does not affect the format of the input.

thomasstvr:
This will close #639

### DIFF
--- a/tests/recipes/wrangles/test_main.py
+++ b/tests/recipes/wrangles/test_main.py
@@ -2492,6 +2492,78 @@ class TestReplace:
         )
         assert df.empty and df.columns.to_list() == ['Abbrev', 'output column']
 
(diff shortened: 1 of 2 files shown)


ebhills (changes requested):
should only work on strings & numbers. Everything else (list, dict, bools, ...) will pass through. Need to add warning (only 1 per type) that states mixed types that are not being processed. Extra credit: capture the row # for the data type issues.

PR #718, opened 2025-05-16
Formatting

thomasstvr:
First round of file formatting


thomasstvr:
Closing this so we don't get bombarded with emails while switching to xlsxwriter

PR #744, opened 2025-07-23
Update extract codes with new parameters

cerdamario13:
Update extract.codes schema to include the latest parameters in recent [release](https://github.com/wrangleworks/Wrangles-Extract-Codes/releases/tag/v2.0.0)

Tested the schema by running generate_recipe_schema.py and adding generated schema file to settings

### DIFF
--- a/tests/recipes/wrangles/test_extract.py
+++ b/tests/recipes/wrangles/test_extract.py
@@ -810,7 +810,7 @@ class TestExtractCodes:
 
     def test_extract_codes_sort_and_max_length(self):
         """
-        Test extract codes with a sort order of shortest
+        Test extract codes with a sort order and min_length
         """
         data = pd.DataFrame({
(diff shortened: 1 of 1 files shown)


thomasstvr:
Might be a good idea to add tests for errors when passing values that are not allowed for these new parameters:

```
ValueError: extract.codes - Status Code: 400 - Bad Request. {"message": "Invalid parameter sort_order. Expected longest or shortest."}
```

Should also add tests using these in combination with each other.

cerdamario13:
> Might be a good idea to add tests for errors when passing values that are not allowed for these new parameters:
> 
> ```
> ValueError: extract.codes - Status Code: 400 - Bad Request. {"message": "Invalid parameter sort_order. Expected longest or shortest."}
> ```
> 
> Should also add tests using these in combination with each other.

Added tests with wrong params and a mix of params used

thomasstvr (approved):

PR #757, opened 2025-08-20
Extract first element

thomasstvr:
Brought this back up to date from several years ago. Basically just brought up to date and dropped first_element from extract.brackets because that is the only one that is meant to return a string.


cerdamario13 (changes requested):
In address, do we need to add the first_element parameter?

```python
def address(
    df: _pd.DataFrame,
    input: _Union[str, int, list],
    output: _Union[str, list],
    dataType: str,
# Add param here???
    **kwargs
) -> _pd.DataFrame:
    """
    type: object
    description: Extract parts of addresses. Requires WrangleWorks Account.
    required:
      - input
      - output
    properties:
      input:
        type:
          - string
          - integer
          - array
        description: Name of the input column.
      output:
        type:
          - string
          - array
        description: Name of the output column.
      dataType:
        type: string
        description: Specific part of the address to extract
        enum:
          - streets
          - cities
          - regions
          - countries
      first_element:
        type: boolean
        description: Get the first element from results
```

Same for html wrangle.

Also the copilot suggestions are worth taking a look since there may not be a [0] when there is no regex match

thomasstvr:
Updated copilot's suggestions and added tests for each case.

ebhills (approved):
Looks good.  thanks

cerdamario13 (approved):

Issue #695, opened 2025-04-09
Copy Breaks With Multiple Columns and One Duplicate


thomasstvr:
When running a copy wrangle with multiple columns and one of the input columns is repeated, a value error is raised:

```
ValueError: copy - Input and output must be the same length
```

This works:
```
wrangles:
  - copy:
    input:
      - col1
      - col1
    output:
      - col1-copy
      - col1-copy-again
```

But this does not:
```
        wrangles:
        - copy:
            input:
                - col1
                - col2
                - col1
            output:
                - col1-copy
                - col2-copy
                - col1-copy-again
```

PR #767, opened 2025-09-03
Copy multi column (Wildcard Bug)

thomasstvr:
This actually ended up being a wildcard expansion issue, not a copy issue. Essentially added gibberish to the column name if it already exists, gibberish plus a counter if that exists, then split on gibberish on the return.

To me, having the same column input multiple times only makes sense on the copy wrangle. Therefore, I only made tests for copy. I can add tests for other wrangles if needed. 

This pr will close issue #695

### DIFF
--- a/tests/recipes/wrangles/test_pandas.py
+++ b/tests/recipes/wrangles/test_pandas.py
@@ -272,6 +272,54 @@ class TestCopy:
         """
         df = wrangles.recipe.run(recipe, dataframe=data)
         assert list(df.columns) == ['col', 'col2', 'col1-copy', 'col2-copy']
+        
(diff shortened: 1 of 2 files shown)


cerdamario13 (approved):
Looks good. 
One thing that I noticed is that when a column is a number it fails. but this issue does not originate form this change.

Also the tests fail of the column name is the same as the gibberish, but what are the odds of that happening?

thomasstvr:
@cerdamario13 yeah exactly, if the column name were the same as the gibberish then it would break but like you said those odds are very low. I forget exactly where, but this method has been used elsewhere in our code, I just made up a new suffix.

PR #782, opened 2025-09-15
Python variable redeux

thomasstvr:


ChrisWRWX (changes requested):
- [ ] Add warning if using invalid vars -> e.g. ${var with spaces}. Allow to work for backwards compatibility and not make this a potential breaking change.
- [x] fix matrix/if - need to merge recipe vars with matrix vars
- [x] move _statement_modifier out of the apply() so it's only executed once.
- [x] merge df_vars into kwargs outside the apply().

PR #784, opened 2025-09-16
Backend query retries

thomasstvr:


ChrisWRWX (changes requested):
- [ ] Get rid of mocked tests
- [x] Don't use retries on POST model/content
- [x] Drop 501 as this isn't likely to change.
- [x] Use new helper function for wrangles.

PR #791, opened 2025-09-22
Standardize spaces

thomasstvr:


cerdamario13:
> > ```python
> > hello\nworld
> > ```
> 
> To me it shouldn't, because technically that does not count as whitespace even though it represents whitespace. Unicode character codes, or escape sequences are not included but instead the space character they represent were along with the \s regex pattern.

Okay sounds good! We can always run a replace on escape sequences if we need them removed

cerdamario13 (approved):

PR #820, opened 2025-11-17
Updated test to reflect new extract properties sorting

thomasstvr:


mborodii-prog (approved):
If it's hot fix, I'm okay with it

PR #834, opened 2025-11-26
813-add-sort-parameter-to-extract-custom-wrangle

mborodii-prog:
This pull request adds support for custom sorting of extracted results in the `extract.custom` wrangle, allowing users to specify how the output should be ordered. It introduces a new `sort` parameter with several sorting options, updates the relevant function signatures, and adds comprehensive tests to ensure correct behavior.


thomasstvr:
- [ ] Add sort type options to the schema as an enum.
- [ ] Add tests using sort and other parameters (use_labels, first_element, case_sensitive, use_spellcheck, where and if), using sort values that are not allowed

thomasstvr (approved):

PR #857, opened 2025-12-31
952-matrix-broken

mborodii-prog:
### Matrix broken fix
PR fixed cases when input variables are not string 
image
## Test recipe:
```
wrangles:
  - matrix:
      variables:
        var: set(col2)
      wrangles:
        - convert.case:
            input: col1
            case: ${var}
            where: col2 = ?
            where_params:
              - ${var}
```

[WranfgleXL isssue link](https://github.com/wrangleworks/WranglesXL/issues/952)


thomasstvr:
Just need to move test from connectors/test_matrix.py to TestMatrix class in recipes/wrangles/test_main.py as this is using the matrix wrangle, not the connector.

mborodii-prog:
@thomasstvr I moved test to the correct place

thomasstvr (approved):

PR #875, opened 2026-01-16
convert.to_json on scalar should product array

mborodii-prog:
# convert.to_json on scalar should product array

##Summary
This PR updates the convert.to_json wrangle to always wrap scalar values in a JSON array, ensuring consistent output for downstream wrangles and simplifying recipe authoring. 
Previously, converting a scalar column to JSON would result in a quoted string, not an array, which was inconsistent with the expectations of other wrangles. This update standardizes the output format and removes the need to use to_list before to_json  for scalar columns.

### What Changed
convert.to_json: Now wraps scalar values (e.g., strings, numbers) in a list before serializing to JSON. For example, "abcd" becomes ["abcd"] instead of "abcd".
### Tests:
Added tests for scalar string, numeric, and multiple columns to verify the new array-wrapping behavior.
Added tests for mixed scalar and list columns.
All existing tests updated and verified to pass with the new logic.

## Example Usage
 **Before:** 

```
wrangles:
  - convert.to_json:
      input: Test
      output: Test_json
# Output: "abcd"
```
**After:**

```
wrangles:
  - convert.to_json:
      input: Test
      output: Test_json

# Output: ["abcd"]
```

**Multiple columns:**

```
wrangles:
  - convert.to_json:
      input:
        - col1
        - col2
      output:
        - col1_json
        - col2_json
# Output: col1_json = ["a"], col2_json = ["b"]
```

No impact on columns that are already lists, dicts, or tuples.

### DIFF
--- a/wrangles/recipe_wrangles/convert.py
+++ b/wrangles/recipe_wrangles/convert.py
@@ -448,22 +448,21 @@ def to_json(
     
     # Ensure input and output are equal lengths
     if len(input) == len(output):
-      # Loop through and apply for all columns
-      for input_column, output_column in zip(input, output):
-        def wrap_scalar(val):
-          # If already a list, dict, or tuple, leave as is; else wrap in list
-          if isinstance(val, (list, dict, tuple)):
-            return val
-          return [val]
-        df[output_column] = [
-          _json.dumps(
-            wrap_scalar(row),
-            ensure_ascii=ensure_ascii,
-            default=handle_unusual_datatypes,
-            **kwargs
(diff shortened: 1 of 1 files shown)


ebhills (dismissed):

thomasstvr (changes requested):
Correct failing tests and could you also fix the indentation on your changes so that it matches the rest of the code with 4 spaces?

mborodii-prog:
@ebhills @thomasstvr Now it's ready for review

PR #879, opened 2026-01-19
876-trainlookup-does-not-catch-no-matching-columns-for-semantic

mborodii-prog:
# Fix: Ensure Semantic Lookup Requires Key or Embeddings Columns

## Summary
This PR addresses an issue where training a semantic (embedding) lookup model would succeed even if neither a "Key" column nor embeddings_columns were provided, resulting in a broken model. The following changes have been made:

## Changes
Validation Added:

In train.lookup, a check now ensures that for semantic/embedding lookups, either a "Key" column or embeddings_columns must be provided.
If neither is present, a clear ValyeError raised:
"Semantic lookup: You must provide either a 'Key' column or 'embeddings_columns' in settings."

**Tests Added:**

A test to confirm that an error is raised if both "Key" and embeddings_columns are missing.
A test to confirm that training succeeds when embeddings_columns is provided.
A test to confirm that training succeeds with[variant: embedding](https://probable-journey-vpgjg6x9g672vv5.github.dev/) and embeddings_columns (no error catching).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,4 +17,4 @@ boto3
 jinja2
 polars==1.33.0
 xlsxwriter
-pydantic
+pydantic>=2.12.0
--- a/tests/connectors/test_train.py
+++ b/tests/connectors/test_train.py
@@ -1,5 +1,7 @@
 import uuid
 
+from pytest_mock import mocker
+
 import wrangles
 import pandas as pd
 import pytest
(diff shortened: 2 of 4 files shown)


tests/test_wrangles.py
```diff
@@ -288,6 +288,49 @@ def test_lookup_list_value_list_column():
...
+    """
+    df = pd.DataFrame({
+            'Value': ['Blade Runner', 'Westworld', 'Interstellar'],
+            'Description': ['Movie 1', 'Movie 2', 'Movie 3']
+    })
+    result = wrangles.recipe.run(
+            """
+            write:
+                - train.lookup:
+                        name: MySemanticLookup
```
thomasstvr: Instead of training a new model with name, can you train an existing model with model_id?

thomasstvr (changes requested):

PR #908, opened 2026-02-18
Fix lookup duplicate issue

### LINKED ISSUE #894: Action broken with multi-column match/no key column
Only overwrite works when retraining a lookup with multiple match columns and no Key column. Insert and upsert simply append the new data, even when they are duplicates of existing data, while update fails completely.

```python
recipe = f"""
write:
    - train.lookup:
        model_id: 1ac1dddb-fcb5-45f3 # Test Multi-Key
        action: insert
        settings:
        variant: semantic
        MatchingColumns:
            - City
            - Country
"""
data = pd.DataFrame({
    'City': ['London', 'Manchester', 'Berlin', 'Tokyo', 'Washington'],
    'Country': ['United Kingdom', 'United Kingdom', 'Germany', 'Japan', 'United States'],
})
df = wrangles.recipe.run(recipe, dataframe=data)
```

Please clone that model and use a different model_id when testing.

mborodii-prog:
This pull request introduces significant improvements to the handling of semantic (embedding-based) lookup models in the WranglesPY data transformation framework.
 The main focus is on supporting multi-column deduplication and update logic for semantic variants using a new `MatchingColumns` parameter, ensuring data integrity and preventing duplicate entries. Additionally, comprehensive tests have been added to validate these behaviors. The pull request also documents the responsibilities, guardrails, and standard operating procedures for the new Lookup Dev Assistant agent.

**Enhancements to semantic lookup model handling:**

* Added support for deduplication and row updates in semantic (non-key) variants using a `MatchingColumns` parameter, ensuring that INSERT and UPSERT actions do not introduce duplicate rows and that UPDATE actions modify only the intended rows. 
* Improved error handling and validation for `MatchingColumns`, including checks for missing columns in both existing and new data, and informative error messages when requirements are not met. 
**Testing improvements:**

* Added comprehensive tests for semantic variant INSERT, UPSERT, and UPDATE actions with multiple matching columns, using mocking to verify correct deduplication and update logic without affecting real models. 

**Documentation and process:**

* Introduced `.github/agents/lookup.agent.md`, documenting the Lookup Dev Assistant agent's responsibilities, coding conventions, quality gates, and standard operating procedures for managing lookup models within WranglesPY.


wrangles/connectors/train.py
```diff
@@ -350,9 +350,44 @@ def _set_variant(mid: str, var: str) -> str:
...
+                        updated = 0
+                        for _, row in df.iterrows():
+                            mask = _pd.Series([True] * len(merged_df))
+                            for col in matching_cols:
+                                mask &= (merged_df[col] == row[col])
+                            if mask.any():
+                                for col in df.columns:
+                                    if col not in matching_cols and col in merged_df.columns:
+                                        merged_df.loc[mask, col] = row[col]
+                                updated += int(mask.sum())
```
thomasstvr: When no values are changed, updated shows that all have been updated when it should show zero.

thomasstvr (changes requested):
There shouldn't be any reason that we have to mock these tests. Let's change them over from being mocked and ensure that we are covering all bases with tests. Overwrite *should* be well tested since that is the default, so it is the others that I am worried about.

- insert, update, upsert with one MatchingColumn (listed as a string)
- insert, update, upsert with multiple MatchingColumns
- insert, update, upsert with a Key column and no MatchingColumns passed

Some of these tests should include variant and some should leave it out. 

I'm sure I'm missing some, or listed some that are already there, but the idea is to add a test for every possible combination.

tests/connectors/test_train.py
```diff
@@ -1058,6 +1059,268 @@ def test_lookup_matchingcolumns_string_not_list(self):
         })
         with pytest.raises(ValueError, match="MatchingColumns.*Not City"):
             wrangles.recipe.run(recipe, dataframe=data)
+
+    def test_lookup_update_matchingcolumns_list(self, caplog):
```
thomasstvr: Duplicate name, slightly different test.

thomasstvr (commented):
Looks great other than the one test name.

thomasstvr (dismissed):

thomasstvr (approved):

PR #925, opened 2026-02-27
allow brackets in select.list_element

### LINKED ISSUE #923: Select list.element should allow [ ]s
Currently select.element errros if user enters range with brackets [0:5].  It only works without the brackets, but this is atypical for how the slice is normally expressed in python. Add check to accept either representation.

mborodii-prog:
This pull request adds support for slice notation when selecting elements from lists using the `wrangles.select.list_element` function, and introduces comprehensive tests to ensure correct behavior with various slice formats.

Enhancements to list element selection:

* Updated `wrangles/select.py` to allow slice notation (including bracketed formats like `[1:3]` and `[::2]`) when specifying indices for `select.list_element`, by stripping brackets before processing the slice.

Testing improvements:

* Added multiple tests in `tests/test_wrangles.py` to verify that `select.list_element` correctly handles slices, bracketed slices, and step values, ensuring robust coverage for the new slice functionality.
usage:
image
image
image
image
image


thomasstvr (changes requested):
Can you also add tests that test this in a recipe in test_select.py? Tests like what is shown in the issue should be fine.

thomasstvr (approved):

PR #931, opened 2026-03-04
New Search, Score , and Retrieve Wrangles for PRA

ebhills:
3 new wrangles with more to come.

- search.find_links

- compute.score_search_results: scoring algo tied directly to find_links response

- search.retrieve_link_content: tied to Google's URL Context API to retrieve content from specific pages

Adding this to main as beta functionality for WrWx only to use


thomasstvr:
This PR is backwards (main into search-web-connector). Please re-submit.

PR #938, opened 2026-03-15
remove_words enhancement

### LINKED ISSUE #895: remove_words enhancements
Current behavior of remove_words function needs changes:

Requirements:

- [ ] tokenize_to_remove should default to true.  change in the code as well as the schema doc string. Also more than one space should be reduced to a single space so that the tokens do not have spaces in them. If the inputs are already lists, then the tokenize step should be ignored
- [ ] Add new parameter: characters_to_consider: all | letters_numbers_only  (default). Default behavior is to remove anything [^a-z0-9] from tokens so that non alphanum characters are not included in the tokens and thus matches occur even when intial strings have miscellaneous punction, /s etc.
- [ ] Regardless of characters_to_consider, any punctuation or spaces at the beginning or end of the input and to_remove values should be removed before tokenization.
- [ ] If either of input or to_remove are lists and the other is just a str, should convert/tokenize the str and run the remove in the list mode.
- [ ] Add defults for ignore_case at True, and note this in the schema
- [ ] Add tests for all combinations.  one example:
       input: Any, Words, Overlap?  
       to_remove: Any Words Overlap
       output: '' (empty string)

mborodii-prog:
### What changed

**New default: `tokenize_to_remove: true`**
String `to_remove` values are now split by whitespace before matching by default, so `'Metal Carbon'` removes `Metal` and `Carbon` individually. Add `tokenize_to_remove: false` to any existing recipes that rely on whole-phrase matching.

**New parameter: `characters_to_consider`** (`letters_numbers_only` | `all`, default `letters_numbers_only`)
Before comparing tokens, non-alphanumeric characters are stripped. This means `A-100` matches `A100`, `Any,` matches `Any`, `Overlap?` matches `Overlap`, etc. — without requiring the input data to be pre-cleaned. Use `characters_to_consider: all` to revert to exact-character matching.

**Leading/trailing punctuation stripped before tokenizing**
Applied to both `input` and `to_remove` values before they are split, regardless of `characters_to_consider`. Handles common noise like trailing `?`, leading `$`, or extra whitespace at string boundaries.

**Multiple consecutive spaces collapsed before tokenizing**
Prevents empty or space-containing tokens when `to_remove` values have inconsistent spacing.

**Lists skip the tokenize step**
If a `to_remove` column already contains a list, each element is treated as an individual token — it is not re-split by whitespace. Likewise for list `input` values. This makes list and string inputs behave symmetrically: either way, you get one token per element.

**`ignore_case` default (`true`) now documented in schema.**

---

### Migration

| Scenario | Action |
|---|---|
| Used `tokenize_to_remove: true` explicitly | None |
| Used `tokenize_to_remove: false` explicitly | None |
| Relied on `tokenize_to_remove` defaulting to `false` (phrase matching) | Add `tokenize_to_remove: false` |
| Need `A-100` to NOT match `A100` | Add `characters_to_consider: all` |

---

### Files changed
- `wrangles/extract.py` — rewrote `remove_words` core logic; added token-based mode with `characters_to_consider` and boundary-stripping; phrase-based regex mode preserved under `tokenize_to_remove: false`
- `wrangles/recipe_wrangles/main.py` — updated defaults and schema; added `characters_to_consider` parameter
- `tests/recipes/wrangles/test_main.py` — updated 3 tests to use explicit `tokenize_to_remove` to preserve their intent; added 13 new tests covering all parameter combinations


ebhills (changes requested):
Summary looks spot on. But the description in the python doc strings and JSON Schema needs to be revised.

thomasstvr:
This looks like it is going a bit too far to me. Automatically reducing whitespace, converting lists to strings, and dropping punctuation are things that should be left up to the user. Doing these automatically leaves them no choice.

There is a lot to unpack here, but overall it seems to be doing too much.

tests/recipes/wrangles/test_main.py
```diff
@@ -1304,6 +1304,465 @@ def test_remove_words_empty(self):
...
+                to_remove:
+                    - to_remove
+                output: output
+        """
+        result = wrangles.recipe.run(recipe, dataframe=data)
+        # Should remove punctuation/spaces from each to_remove element
+        assert result['output'][0].strip() == ''
+        assert result['output'][1].strip() == ''
+
+    def test_remove_words_list_input_str_to_remove(self):
```
thomasstvr: Multiple tests with the same name

thomasstvr (changes requested):

thomasstvr:
So a couple things that are going to cause issues.

tokenize_to_remove defaulting to true instead of false is going to break recipes that are removing things like "Bearing Co." from something like "SKF 6202 ball bearing".

Also, removing punctuation from the input will cause issues down the line in the recipe. For example, in the "1b. Refine Input for RT PRA" recipe, MPN is removed from the description early on in the recipe. This causes parenthesis to be dropped at the end of the description, which are important for things like extract.brackets (used later in the recipe). 

There may or may not be more, but these are the two big things I noticed with a small amount of testing.

To be clear, I am not opposed to these changes. I do think they should be opt-in though instead of op-out in order to keep from breaking recipes already in use.

ebhills:
Queue triage (2026-07-27)

- **Next action:** Address the requested behavior changes for issue #895, update from `main`, and re-request one reviewer. Keep it in v1.20 only if it can meet the new freeze.

PR #939, opened 2026-03-15
Move optional packages to separate requirements file

### LINKED ISSUE #913: move large, optional packages to separate add-ons package
A number of our imports are not being used.  For those that are therefore optional, we should put them into a seprate package so that PY core is smaller / faster install.

mborodii-prog:
This pull request introduces a new requirements file for full dependency management and updates a test assertion to match the expected translation output. The main changes are grouped into dependency management and test adjustments.

Dependency management:

* Added a new `requirements-full.txt` file to specify all optional dependencies and connectors, enabling easier installation for running the complete test suite.

Test adjustments:

* Updated the assertion in `test_translate` and `test_translate_list` in `tests/test_wrangles.py` to match the expected German translation output without a period.


ebhills (commented):
- [x] Let's discuss where we draw the line between these 2 installs.  If I follow the changes, it seems to me that we have taken too much out of the baseline configuration.  It should support 80% of our capabilities.  The motivation for this effort was primarily to leave out database connectors and other rarely used packages needed to support edge case functionality.
- [x] In order to make the assessment, please summarize what functionality is supported by base packages and what additional capabilities are unlocked with -full?  Once this is decided, it should be explained in the .txt files, the repo README.md, and our wrangles.io install.
- [x] Requirements.txt should have the grouping headers just like the -full.
- [x] Need to add packages for serpapi, google genai, and openai sdk(s)

tests/test_wrangles.py
```diff
@@ -843,4 +843,4 @@ def test_compare_overlap_exact_match_custom():
     Test compare.overlap with exact_match parameter
     """
     result = wrangles.compare.overlap([['test', 'test']], exact_match='MATCH')
-    assert result == ['MATCH']
\ No newline at end of file
+    assert result == ['MATCH']
```
thomasstvr: Please drop this change just for cleanliness.

thomasstvr (changes requested):
One tiny little thing to clean up, but looks good otherwise.

thomasstvr (approved):

PR #940, opened 2026-03-16
Read/Write Meta Data With Train Connector

### LINKED ISSUE #887: Read/Write Meta Data With Train Connector
The meta data is already stored, and should be accessible from the response. The train connector should be configured to be able to read and write meta data to the model. This may be some other variant such as train.meta_data or maybe just added to the individual variants such as train.lookup.

mborodii-prog:
# Add `train.meta_data` Connector

## Summary

Adds a new `train.meta_data` read/write connector that exposes model metadata through the standard recipe interface. Previously, metadata was only accessible programmatically via `wrangles.data.model()`. Now it can be read and updated directly in a recipe, including as part of a read → wrangle → write pipeline.

---

### New: `train.meta_data` connector (`wrangles/connectors/train.py`)

**Read** — returns a model's metadata as a single-row DataFrame:

```yaml
read:
  - train.meta_data:
      model_id: abc123-xxxx
```

Returns columns like `name`, `variant`, `purpose`, `batch_size`, `settings`.

**Write** — updates metadata fields for an existing model (partial update — only fields provided are changed):

```yaml
write:
  - train.meta_data:
      model_id: abc123-xxxx
```

The DataFrame passed in determines which fields are updated. `NaN` values are ignored.

---

### New: `data.model_update()` (`wrangles/data.py`)

Low-level function that calls `PUT /model/metadata` with a dict of fields to update.

```python
wrangles.data.model_update(model_id, {"name": "New Name", "batch_size": 500})
```

---

### Read metadata
```yaml
read:
  - train.meta_data:
      model_id: abc123-xxxx
```

| name | variant | purpose | batch_size | settings |
|------|---------|---------|------------|---------|
| My Lookup | key | lookup | None | {} |

---

### Update the model name
```yaml
write:
  - train.meta_data:
      model_id: abc123-xxxx
```
Input DataFrame:
| name |
|------|
| My Renamed Lookup |

---

### Read → modify → write back in one recipe
```yaml
read:
  - train.meta_data:
      model_id: abc123-xxxx

wrangles:
  - convert.case:
      input: name
      case: upper

write:
  - train.meta_data:
      model_id: abc123-xxxx
```

---

## Tests Added (`tests/connectors/test_train.py`)

| Test | Type | Description |
|------|------|-------------|
| `test_meta_data_read` | mocked | Fields returned correctly from API response |
| `test_meta_data_read_all_fields` | mocked | All columns present including nested `settings` |
| `test_meta_data_write` | mocked | `model_update` called with correct model_id and payload |
| `test_meta_data_write_ignores_nan` | mocked | NaN values stripped before sending to API |
| `test_meta_data_write_access_denied` | mocked | RuntimeError propagates on 401/403 |
| `test_meta_data_read_integration` | integration | Real API call — checks expected fields exist |
| `test_meta_data_write_integration` | integration | Real API call — updates name and verifies change |

Integration tests use a shared `module`-scoped fixture that creates a temporary lookup model once for the test run.

---

## Files Changed

| File | Change |
|------|--------|
| `wrangles/connectors/train.py` | Added `meta_data` class with `read()` and `write()` |
| `wrangles/data.py` | Added `model_update()` function |
| `tests/connectors/test_train.py` | Added 7 tests (5 mocked, 2 integration) |
| `train_meta_data_demo.ipynb` | Notebook demo with end-to-end examples |


ebhills (commented):
Several additional requirements come to mind:

- [x] I was expecting we would get all of the metadata that is visible on the Details Card. This would include last used/created/updated dates, Tags, and Notes.
- [x] should the schema for read list the output columns available? I think it would be helpful to document what will be coming back in the response.
- [x] Logging info -- on write, let's add a summary statement that says 'Name was updated to My New Name' etc.
- [ ] Can we expand the read interface to include reading multiple models in one call? Maybe we could use Vlad's existing XL filter logic to select based on Type / Purpose. Would be good to add Permissions as filter criteria as well (eg where Org=WrangleWorks & Role=Admin).  If this enhancement will take a long time, let's make it into a separate issue.

thomasstvr (changes requested):
- [x] Add error handling for wrong data types being passed. For example, a string passed for batch_size.
- [x] Only allow certain fields to be written. Lets only use name, batch_size, tags, notes, and settings for now, with error handling for anything else.

ebhills (commented):
1. Please add the list of allowed fields to the write schema
2. The test all fields appears to be missing many from the expanded field set

ebhills (approved):

thomasstvr:
@ebhills like you mentioned, let's plan on expanding this with the multiple model thing later. I could see that being a bit complicated.

Issue created #962, feel free to add to it.

thomasstvr (approved):

PR #957, opened 2026-04-06
Create Column should NOT error if column exists

mborodii-prog:
This pull request updates the behavior of the `create.column` wrangle to provide more flexible handling when attempting to create columns that already exist in the DataFrame. Instead of always raising an error, the function now supports a new `value_if_exists` parameter that controls whether to leave the column unchanged, coalesce (fill only empty/null cells), or overwrite the column. The tests have been updated to verify these scenarios.

**Enhancements to `create.column` behavior:**

* Added a new `value_if_exists` parameter to the `column` function in `wrangles/recipe_wrangles/create.py`, allowing users to specify how to handle existing columns: `'existing'` (default, leaves column unchanged), `'coalesce'` (fills only empty/null cells), or `'new'` (overwrites the column).
*  Added new coalesce_value parameter ('existing' default / 'new'), implementing ebhills' decision: when value_if_exists: coalesce, you can now choose whether the
  existing value or the new value takes priority on non-empty conflicts, with the other side only filling in when the primary is empty/null. 
* Updated the implementation of the `column` function to support the new behaviors, removing the previous error-raising logic and handling each mode accordingly.

**Test updates:**

* Modified and added tests in `tests/recipes/wrangles/test_create.py` to check the default behavior (no error, column unchanged), and the new behaviors for `value_if_exists: existing`, `value_if_exists: new`, and `value_if_exists: coalesce`.

**Examples of usage**

* The column already exists — do nothing, raise no error.
This is the default so you can omit the parameter entirely.
```
wrangles.recipe.run(
    recipe="""
    wrangles:
      - create.column:
          output: status
          value: unknown
          # value_if_exists: existing  <-- this is the default
    """,
    dataframe=data.copy()
)
```

```
df = wrangles.recipe.run(
    recipe="""
    wrangles:
      - create.column:
          output:
            - status      # already exists → skipped
            - region      # new → created
          value: EMEA
    """,
    dataframe=data.copy()
)
```
*Fill gaps (`null` and empty string) in the existing column while preserving
rows that already have a value.
```
 wrangles.recipe.run(
    recipe="""
    wrangles:
      - create.column:
          output: status
          value: unknown
          value_if_exists: coalesce
    """,
    dataframe=data.copy()
)
```

*Unconditionally overwrite every cell in the existing column.

```
wrangles.recipe.run(
    recipe="""
    wrangles:
      - create.column:
          output: status
          value: reset
          value_if_exists: new
    """,
    dataframe=data.copy()
)
```

### DIFF
--- a/tests/recipes/wrangles/test_create.py
+++ b/tests/recipes/wrangles/test_create.py
@@ -92,44 +92,83 @@ class TestCreateColumn:
         df = wrangles.recipe.run(recipe, dataframe=data)
         assert df.iloc[0]['column3'] in [True, False]
         
-    def test_column_exists(self):
+    def test_column_exists_no_error(self):
         """
-        Check error if trying to create a column that already exists
+        Default behaviour: creating a column that already exists should not raise
+        an error — the existing column is left unchanged.
         """
-        data = pd.DataFrame({
-        'col': ['data1']
-        })
+        data = pd.DataFrame({'col': ['data1']})
         recipe = """
         wrangles:
         - create.column:
             output: col
+            value: new_value
         """
-        with pytest.raises(ValueError) as info:
-            wrangles.recipe.run(recipe, dataframe=data)
-        assert (
-            info.typename == 'ValueError' and
-            '"col" column already exists in dataFrame.' in info.value.args[0]
-        )
+        df = wrangles.recipe.run(recipe, dataframe=data)
+        assert df['col'][0] == 'data1'
 
-    def test_column_exists_list(self):
+    def test_column_exists_list_no_error(self):
         """
-        Check error if trying to create a list of columns where one already exists
+        Default behaviour: creating a list of columns where one already exists
+        should not raise an error — the existing column is left unchanged.
         """
-        data = pd.DataFrame({
-        'col': ['data1']
-        })
+        data = pd.DataFrame({'col': ['data1']})
         recipe = """
         wrangles:
         - create.column:
             output:
                 - col
+                - col2
+            value: new_value
         """
-        with pytest.raises(ValueError) as info:
-            wrangles.recipe.run(recipe, dataframe=data)
-        assert (
-            info.typename == 'ValueError' and
-            "['col'] column(s)" in info.value.args[0]
-        )
+        df = wrangles.recipe.run(recipe, dataframe=data)
+        assert df['col'][0] == 'data1' and df['col2'][0] == 'new_value'
+
+    def test_column_exists_value_if_exists_existing(self):
+        """
+        Explicit value_if_exists: existing leaves the column unchanged.
+        """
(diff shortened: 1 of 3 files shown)


thomasstvr:
Which way does this coalesce? New into existing or existing into new? Should we allow both?

@ebhills any opinion on what we call the new value_if_exists parameter? Here is the current schema for it:

```
      value_if_exists:
        type: string
        description: >-
          Determines behaviour when the output column already exists.
          existing (default): leave the column unchanged.
          coalesce: fill empty/null cells with the new value, keeping non-null cells.
          new: overwrite the entire column with the new value.
        enum:
          - existing
          - coalesce
          - new
```

ebhills:
@thomasstvr - good question re coalesce. I envisioned the existing values would be treated as colm 1 (keep if exists) and then new value would just fill when empty. I suppose we could let user decide with another param coalesce_value: prefer existing / prefer new

mborodii-prog:
@ebhills let me know if this notebook is up to date your expected behavior:  
[demo_issue_956.ipynb](https://github.com/user-attachments/files/28973037/demo_issue_956.ipynb)

ebhills:
there are 2 unresolved issues from copilot. if they ahve been fixed then mark them resolved.  also, I don't see where we decided which direct coalesce goes in.  I assume it would be keep original value if exists, only filling in blanks with (new) value.

ebhills:
I think you should paste the wrangles schema here as a comment in the Conversation (when it changes). That will make it very easy for us to understand how it works and test it.

mborodii-prog:
@ebhills  Direction is settled — added a coalesce_value parameter ('existing' default / 'new') alongside value_if_exists. Default behavior matches what you described: keep the original value if it exists, only filling blanks (NaN/None, and empty string for object-dtype columns) with the new value. Setting coalesce_value: new flips priority so the new value wins on conflicts, falling back to the existing value only when the new value is empty/null.

value_if_exists:
    type: string
    description: >-
      Determines behaviour when the output column already exists.
      existing (default): leave the column unchanged.
      coalesce: fill empty/null cells with the new value, keeping non-null cells.
      new: overwrite the entire column with the new value.
    enum:
      - existing
      - coalesce
      - new
  coalesce_value:
    type: string
    description: >-
      Only used when value_if_exists is coalesce. Determines which side
      is preferred when both the existing and new values are non-empty.
      existing (default): keep the existing value, only fill empty/null cells with the new value.
      new: keep the new value, only fall back to the existing value where the new value is empty/null.
    enum:
      - existing
      - new

  This matches the live docstring in wrangles/recipe_wrangles/create.py:115-135. Let me know if you want me to also double check tests cover the coalesce_value: new path before you push anything else.

ebhills (approved):

PR #977, opened 2026-04-28
New Model Training Refactor

### LINKED ISSUE #972: [BUG] Train.extract trains broken models
**Describe the bug**
When training a new extract wrangle using name, the connector trains a broken model. The model does not show any error, but errors with "Something went wrong." when run. 

**To Reproduce**
1. Train new extract wangle (using name param).
2. Try to use new model in excel or wranglespy

**Expected**
Should create a working model

**Additional context**
This is probably best done in VS code to make it easier to use your wrangles log in info, otherwise the models will be created on the service user's account.

**Acceptance criteria**
- [x] The new models work

mborodii-prog:
This pull request fixes a bug (Bug #972) where models created using the `name` parameter for various training types (`classify`, `extract`, `lookup`,
  and `standardize`) could be created in a broken state, causing errors on inference immediately afterward. The root cause was a malformed model-creation
  request (e.g. missing `variant` for non-default extract variants) rather than backend eventual-consistency, so the fix corrects the create payload and
  consolidates duplicated logic — it does not add client-side polling/retries on creation.

  ### Bug Fixes

  * Introduced `_create_model_with_content`, a single shared helper used by `classify`, `extract`, `lookup`, and `standardize` for model creation via
  `name`. It issues one POST to `/model/content` with the full training payload and correct params (including `variant` for extract), and parses/logs the
  new `model_id` from the response.
  * This removes the previous per-type duplication of the POST + model-id-parsing logic, so all four training types stay in sync going forward.

  ### Test Enhancements

  * Added integration tests in `tests/connectors/test_train.py` verifying that models created via `name` work correctly, including an end-to-end
  regression test (`test_extract_name_creates_working_model_after_short_wait`) that trains an extract model by name and confirms it produces correct
  inference results shortly after creation — validating the fix without relying on a retry/polling mechanism.
  * Added a `_delete_model` test helper to clean up models created during tests.

  ### Logging

  * Model-id logging on creation is now centralized in `_create_model_with_content`, with a fallback log message if the id can't be parsed from the
  response.

### DIFF
--- a/tests/connectors/test_train.py
+++ b/tests/connectors/test_train.py
@@ -22,7 +22,7 @@ def _delete_model(model_id):
     except Exception:
         pass
 
-def _wait_for_model(recipe, dataframe=None, max_wait=60, interval=5):
+def _wait_for_model(recipe, dataframe=None, max_wait=120, interval=5):
     """
     Newly created models compile asynchronously after the PUT.
     Retry until:
--- a/tests/connectors/test_train.py
+++ b/tests/connectors/test_train.py
@@ -1,3 +1,4 @@
+from tabnanny import check
 import uuid
 import requests as _requests
 from pytest_mock import mocker
@@ -20,6 +21,29 @@ def _delete_model(model_id):
         )
     except Exception:
         pass
+
+def _wait_for_model(recipe, dataframe=None, max_wait=60, interval=5):
+    """
+    Newly created models compile asynchronously after the PUT.
+    Retry until:
+      - the recipe raises no exception, OR
+      - max_wait seconds elapse.
+    Some model types (standardize, classify, lookup) return the input unchanged instead
+    of raising an exception when not ready — use `check` to detect that case.
+    """
+    import time
+    deadline = time.time() + max_wait
+    last_err = None
+    while time.time() < deadline:
+        try:
+            result = wrangles.recipe.run(recipe, dataframe=dataframe)
+            return result
+        except Exception as e:
+            last_err = e
+        time.sleep(interval)
+    if last_err:
+        raise last_err
+    raise AssertionError(f'Model did not produce expected output within {max_wait}s')
 
 class LogCapture(logging.Handler):
     def __init__(self, *args, **kwargs):
@@ -187,9 +211,8 @@ def test_classify_name_creates_working_model(caplog):
             break
 
     assert new_model_id is not None, 'model_id was not logged after training'
-    time.sleep(3)
 
-    result = wrangles.recipe.run(
+    result = _wait_for_model(
         f"""
         wrangles:
             - classify:
@@ -565,9 +588,7 @@ class TestTrainExtract:
 
         assert new_model_id is not None, 'model_id was not logged after training'
 
-        time.sleep(3)
-
-        result = wrangles.recipe.run(
+        result = _wait_for_model(
             f"""
             wrangles:
                 - extract.custom:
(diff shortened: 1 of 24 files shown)


wrangles/train.py
```diff
@@ -35,12 +35,38 @@ def classify(training_data: list, name: str = None, model_id: str = None):
...
             _logging.info(f": Creating new classify model :: {name}")
-            response = _requests.post(
+            create_response = _requests.post(
                         f'{_config.api_host}/model/content',
                         params={'type':'classify', 'name': name},
                         headers={'Authorization': f'Bearer {_auth.get_access_token()}'},
-                        json=training_data
+                        json=training_data,
                     )
+            if not create_response.ok:
```
thomasstvr: Please add comments to any new code. They don't have to be excessive, but it would be nice to explain the reasoning/purpose of sections of code.

tests/connectors/test_train.py
```diff
@@ -148,6 +185,49 @@ def test_classify_write_logs_new_model_id_integration(caplog):
   
     assert any(record.message for record in caplog.records if record.levelname == "INFO" and "New classify model created" in record.message)
 
+def test_classify_name_creates_working_model(caplog):
```
thomasstvr: Please add all train.classify tests to a class for sorting

tests/connectors/test_train.py
```diff
@@ -1,12 +1,49 @@
...
 from pytest_mock import mocker
 
 import wrangles
 import pandas as pd
 import pytest
 import logging
 import re
+import time
+
+def _delete_model(model_id):
```
thomasstvr: Let's plan to move this outside of tests since it would be useful to delete models in general.

thomasstvr (changes requested):

mborodii-prog:
@thomasstvr pls review one more time

thomasstvr:
@mborodii-prog sorry, I meant for us to plan to create a wrangle from that function at a later time, and not in this branch. Please revert the changes.

wrangles/recipe_wrangles/main.py
```diff
@@ -2215,3 +2217,21 @@ def Try(
...
+    :param model_type: The type of the model (classify, extract, lookup, standardize).
+    """
+    params = {'model_id': model_id}
+    if model_type:
+        params['type'] = model_type
+    return _requests.delete(
+        f'{_config.api_host}/model/content',
+        params=params,
+        headers={'Authorization': f'Bearer {_auth.get_access_token()}'},
+    )
```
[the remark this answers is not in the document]
thomasstvr: This is a great point, which I will add to issue #991.

mborodii-prog:
@thomasstvr changes for delete are reverted. also I added enhancement for deduplication code

thomasstvr:
@mborodii-prog something happened with this branch, it looks like it is very far behind main. Please update with main and I will have another look

mborodii-prog:
@thomasstvr branch is fixed

mborodii-prog:
@ebhills @thomasstvr you can use recipe '972 fix demo' for testing in QA.

ebhills:
Queue triage (2026-07-27)

- **Next action:** Address or explicitly answer the outstanding requested changes, then re-request one primary reviewer. This is a near-term v1.20 candidate once the review decision is current.

ebhills (approved):

Issue #980, opened 2026-05-01
[REFACTOR] PRA Cleanup


thomasstvr:
**What needs refactoring**
PRA needs to be gone through and checked for any issues that might pop up, ways to improve etc since it is currently a minimal viable product (mvp).

**Current state**
MVP

**Proposed changes**
More tests to add, and lookover closely for any unused code or general enhancements. The below list is a good starting place, but is neither a definite nor an exhaustive list.

- Is format.retrieved_content_to_text being used?
- Is format.raw_search_results_to_text being used?
- Does format.search_result_to_text need a recipe wrangle version?
- What exactly is web.py doing that couldn't be part of compute.py? Does it need a recipe wrangle version? Does it need tests?
- compute.score_search_results needs much more tests
- Is mask_original_term being used?
- Do compute.mask_original_term, compute.normalize_alphanum, and compute.partial_ratio need recipe wrangle versions?
- Ensure all new wrangles are in alphabetical order within their respective files.
- compute.score_search_results has a lot of parameters and therefore should be tested with a lot of combinations of these parameters. Also, are all parameters needed? Is there an easier way to condense them or use kwargs for easier expansion in the future? Are the parameters described correctly in the schema?
- Should we add the google-genai package to requirements since it is required for retrieve_link_content?
- What is scripts/test_search.py and what is it doing?
- Do all of the examples in recipe_search_web.wrgl.yml work?
- compute.score_search_results slices the input list in a very specific order. These should be turned into individual parameters.
- score_search_results is outputting empty strings for columns with no results. This causes those rows to then be filtered out and lost with no reason why. Rows marked filtered should only be done so for blacklisted words and nothing falsey should pass through otherwise.

**Benefits**
- Production worthy, robust PRA

thomasstvr:
From copilot:

wrangles.__init__ sets search to the wrangles.clients.serp_api module, which masks the new wrangles/search.py public module when users do import wrangles; wrangles.search.... This makes wrangles.search.find_links / retrieve_link_content unavailable from the package root. Export the new wrangles.search module (e.g., from . import search) or re-export specific functions, and avoid aliasing the client module as search.

Suggested change:
```python
from .clients import serp_api as search
``` 
->
```python
from . import search
from .clients import serp_api
```

I tried adding the direct call with tests, which did not work and will need to be added once this is resolved.

thomasstvr:
From copilot regarding wrangles/compute.py:

[...]

PR #981, opened 2026-05-01
Search web connector refactored

thomasstvr:
This is 1/2 pull requests based on branched created from the search-web-connector branch, aka PRA 3.0. This was done in order to break it up into scoring and searching pieces and make it more digestible.


wrangles/__init__.py
```diff
@@ -23,6 +23,7 @@
 from .standardize import standardize
 from . import format
 from . import openai
+from .clients import serp_api as search
```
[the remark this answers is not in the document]
thomasstvr: Valid point, but let's plan to revisit this

tests/recipes/wrangles/test_search.py
```diff
@@ -0,0 +1,537 @@
...
+        recipe = """
+        wrangles:
+          - search.retrieve_link_content:
+              input: summary
+              output:
+                - retrieved_data
+                - Retrieved Content
+              output_format: json
+              api_key: ${GEMINI_API_KEY}
+              threads: 10
```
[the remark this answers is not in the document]
thomasstvr: This is in GitHub secrets

wrangles/clients/gemini.py
```diff
@@ -0,0 +1,170 @@
...
+    """Lazy loader for Google GenAI SDK to prevent hard dependencies."""
+    try:
+        from google import genai
+        from google.genai import types, errors
+        return genai, types, errors
+    except ImportError:
+        raise ImportError(
+            "The google-genai package is required for URL context retrieval. "
+            "Install it with: pip install google-genai"
+        )
```
mborodii-prog: you can use _LazyLoader() against this additional function

requirements.txt
```diff
@@ -17,5 +17,6 @@ boto3
 jinja2
 polars==1.33.0
 xlsxwriter
+serpapi
```
mborodii-prog: I will add specific version here as it  has had breaking API changes between versions

wrangles/recipe_wrangles/search.py
```diff
@@ -0,0 +1,313 @@
...
+
+def retrieve_link_content(
+    df: _pd.DataFrame,
+    input: str | list,
+    output: str | list | None = None,
+    client: str = "google_url_context",
+    api_key: str | None = None,
+    prompt: str | None = None,
+    model_id: str = "models/gemini-3-flash-preview",
+    output_format: str = "json",
```
mborodii-prog, line 183: retrieve_link_content() in wrangles/search.py (the core layer) defaults to output_format: str = "markdown". The recipe-layer wrapper in wrangles/recipe_wrangles/search.py defaults to output_format: str = "json". When a user calls the core function directly (e.g., wrangles.search.retrieve_link_content(...)), they get markdown output. When they call it via a recipe (search.retrieve_link_content), they get JSON. The divergence will cause confusion and may break code that relies on the default behavior.

wrangles/clients/serp_api.py
```diff
@@ -0,0 +1,223 @@
...
+        except ImportError:
+            raise ImportError(
+                "The serpapi package is required for search functionality. "
+                "Install it with: pip install serpapi"
+            )
+
+    def search_single(self, query: str, n_results: int = 5, kwargs: dict = None, query_index: int | None = None) -> dict:
+        """Perform a single web search using SerpAPI."""
+        query_str = str(query).strip().lower()
+        if query is None or not query_str or query_str in ("none", "nan", "nat"):
```
mborodii-prog, line 123: I will add smth like that   _EMPTY_SENTINELS = {"", "none", "null", "nan", "nat"}

  def _is_empty(value) -> bool:
      return value is None or str(value).strip().lower() in _EMPTY_SENTINELS

  Then both checks become if _is_empty(api_key) and if _is_empty(query) — consistent, testable, and no redundancy.

wrangles/clients/serp_api.py
```diff
@@ -0,0 +1,223 @@
...
+            
+            try:
+                price = float(price_str)
+                currency = currency_symbol.upper()
+            except ValueError:
+                pass
+
+    if price is None and not currency and not availability:
+        return {}
+
```
mborodii-prog, line 89: I don't like all this checks. Several issues here:

  1. Asymmetric currency coverage

  C$, CA$, and Mex$ appear in the prefix regex but not the suffix one. A price written 100 C$ or 1500 Mex$ falls through both branches and returns nothing.

  2. currency_symbol.upper() doesn't normalize symbols

  $.upper() → $, not USD. £ → £, not GBP. The output mixes ISO codes (CAD, EUR) with raw symbols ($, £), making the result inconsistent for any downstream consumer
   that expects USD vs $ to mean the same thing.

  3. The decimal heuristic has blind spots

  if re.search(r',\d{2}$', raw_price):   # exactly 2 digits → treat as decimal
      price_str = raw_price.replace(',', '.')
  else:                                    # anything else → strip comma
      price_str = raw_price.replace(',', '')

  - "1,2" — 1 digit, doesn't match → stripped to "12" (wrong, should be 1.2)
  - "1,2345" — 4 digits, doesn't match → stripped to "12345" (wrong)
  - "1,000" — 3 digits, doesn't match → stripped to "1000" (correct by accident)

  The robust rule is: 3 digits after comma = thousands separator, 1–2 digits = decimal. The current check only handles the 2-digit case.

  4. (\d+(?:[.,]\d+)*) is too permissive

  This matches pathological strings like 1.2.3.4 or 1,2,3,4. For currency you'd want something closer to \d{1,3}(?:[.,]\d{3})*(?:[.,]\d{2})?.

  5. Symbol-to-ISO mapping should be a dict, not .upper()

  _SYMBOL_TO_ISO = {"$": "USD", "£": "GBP", "€": "EUR", "C$": "CAD", "CA$": "CAD", "Mex$": "MXN"}
  currency = _SYMBOL_TO_ISO.get(currency_symbol, currency_symbol.upper())

wrangles/clients/serp_api.py
thomasstvr: Can you add this to issue #980?

wrangles/clients/serp_api.py
thomasstvr: Can you add this to issue #980?

mborodii-prog (approved):

PR #984, opened 2026-05-13
implement openAi best practicies

mborodii-prog:
# OpenAI Responses API Integration for `extract.ai`

## Summary

Migrates `extract.ai` from the legacy Chat Completions API (function-calling) to the OpenAI **Responses API** with structured outputs, while preserving full backwards-compatibility for callers using the old endpoint.

---

### `wrangles/extract.py`

- **Default endpoint** switched to `https://api.openai.com/v1/responses`.
- **`strict`** now defaults to `True` (structured output validation always on).
- **`messages`** default changed from `[]` to `None` to avoid the mutable-default-argument pitfall.
- **New parameter `reasoning`** — passes Responses API reasoning options; defaults to `{"effort": "none"}` automatically for GPT-5 models.
- **Routing logic** — if the caller's `url` contains `/chat/completions`, the legacy `chatGPT` code path is used unchanged. All other URLs go through the new Responses API path.
- **Improved system prompt** — more precise instructions to the model: only extract requested fields, return schema-compatible empty values when a field is absent, call `parse_output` exactly once.
- Schema is now sanitized via `openai_responses.sanitize_schema()` before being sent to either API path.

### `wrangles/openai.py` (Chat Completions path)

- Added `required_fields: list` parameter so the caller supplies the list of expected output keys; the function no longer introspects the schema payload to discover them.
- Replaced repetitive inline error dicts with an `_error_result()` helper.
- Added explicit checks for model refusals (`message.refusal`) and unexpected finish reasons.
- Minor style clean-up (consistent keyword spacing, tighter exception handling).

### `wrangles/openai_responses.py` (new module)

New helper module that encapsulates all Responses API concerns:

| Function | Purpose |
|---|---|
| `sanitize_schema()` | Strips unsupported JSON Schema keys; converts `type` arrays to `anyOf` for strict mode |
| `build_response_model()` | Generates a Pydantic model from the output schema for response validation |
| `schema_type_to_python()` | Maps JSON Schema types to Python/Pydantic types (including enums, arrays, nested objects) |
| `call_structured()` | Sends a request to the Responses API, validates the structured JSON response against the Pydantic model, retries with exponential back-off |
| `extract_response_text()` | Parses `output_text` from the Responses API response envelope |
| `prompt_cache_key()` | Generates a stable SHA-256 cache key from namespace + model + schema for prompt caching |
| `format_input_data()` | Serialises dict/list inputs to JSON, scalars to string |
| `error_result()` | Builds a `{field: error_message}` dict for all required output fields |
| `sanitize_request_params()` | Removes params unsupported by the Responses API (e.g. `seed`) |

### `tests/test_wrangles.py`

Added `TestExtractAIResponses` class with six integration tests against the real API:

- `test_string_output` — plain-string prompt output
- `test_single_schema_output` — single-field JSON schema (integer)
- `test_list_input` — list of inputs, multi-field schema
- `test_nested_array_enum_and_boolean` — complex schema with array, enum, boolean, and nested object fields
- `test_messages` — additional instructions via `messages`
- `test_legacy_chat_completions` — verifies the `/chat/completions` backwards-compatibility path still works

---

## Backwards Compatibility

Existing callers that pass `url="https://api.openai.com/v1/chat/completions"` explicitly are routed through the unchanged Chat Completions code path and are unaffected. Callers that relied on the old default URL will automatically use the Responses API.

### DIFF
--- a/tests/test_wrangles.py
+++ b/tests/test_wrangles.py
@@ -420,6 +420,164 @@ def test_extract_ai_properties_list():
     )
     assert isinstance(result, list) and 'unit' in result[0]
 
+
+
+def test_extract_ai_responses_string_output_real_api():
+    result = wrangles.extract.ai(
+        "Paint swatch label: Signal Red RAL 3001.",
+        api_key=os.environ['OPENAI_API_KEY'],
+        output="Return only the color name mentioned in the data.",
+        messages="Use title case and do not include color standards or codes.",
+        threads=1,
+        timeout=60,
+    )
+
+    assert result == "Signal Red"
+
+
+def test_extract_ai_responses_single_schema_output_real_api():
+    result = wrangles.extract.ai(
+        "Shipment: 8 boxes arrived at Dock 4.",
+        api_key=os.environ['OPENAI_API_KEY'],
+        output={
+            "type": "integer",
+            "description": "The number of boxes in the shipment.",
+        },
+        threads=1,
+        timeout=60,
+    )
+
+    assert result == 8
+
+
+def test_extract_ai_responses_list_input_real_api():
+    result = wrangles.extract.ai(
+        [
+            "Badge A belongs to Priya Shah in Finance.",
+            "Badge B belongs to Marco Diaz in Legal.",
+        ],
+        api_key=os.environ['OPENAI_API_KEY'],
+        output={
+            "name": {
+                "type": "string",
+                "description": "The person's full name.",
+            },
+            "department": {
+                "type": "string",
+                "description": "The department name.",
+            },
+        },
+        threads=2,
+        timeout=60,
+    )
+
+    assert result == [
+        {"name": "Priya Shah", "department": "Finance"},
+        {"name": "Marco Diaz", "department": "Legal"},
+    ]
+
+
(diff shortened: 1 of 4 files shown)


tests/test_wrangles.py
```diff
@@ -420,6 +420,164 @@ def test_extract_ai_properties_list():
     )
     assert isinstance(result, list) and 'unit' in result[0]
 
+
+
+def test_extract_ai_responses_string_output_real_api():
```
thomasstvr: Please place new tests into a class for sorting

wrangles/extract.py
```diff
@@ -53,13 +54,14 @@ def ai(
     api_key: str,
     output: dict = None,
     model_id: str = None,
-    model: str = "gpt-4.1-mini",
+    model: str = "gpt-5.5",
```
thomasstvr: There is a massive cost difference between gpt-4.1-mini and gpt-5.5 and we do not want to catch anyone off guard with this. Let's default to gpt-5.4-mini

thomasstvr (changes requested):

PR #988, opened 2026-05-19
Accept path objects in read / write / run recipe

### LINKED ISSUE #986: [ENHANCEMENT] - Accept path objects in read / write / run recipe
**Feature description**
read / write / run wrangles should accept path object. 

**Problem it solves**
I spend waaaayyy too much time pointing recipes and files.  the path library can make it more deterministic but it produces path objects not strings.

**Proposed solution**
add option to provide path ojects in addition to string file names or yaml directly

**Use case**
all the time in PY recipes

**Acceptance criteria**
- [ ] 
- [ ] 
- [ ]

mborodii-prog:
## Summary

`wrangles.recipe.run()` and the file connector `read`/`write` functions now accept
[`pathlib.Path`](https://docs.python.org/3/library/pathlib.html) objects (and any
`os.PathLike`) wherever a file-path string was previously required.

This removes the need to wrap every path in `str()` when building recipes with
`pathlib`, which is the recommended way to handle paths in modern Python.

## Problem

Passing a `Path` object raised `ValueError: Recipe passed in as an invalid type`
because `_load_recipe()` in `recipe.py` tested `isinstance(recipe, str)` first.
A `Path` is not a `str`, so the code fell into the pre-parsed-recipe branch and
tried to YAML-serialise the `Path` object — which always failed.

The file connector had the same issue: `file.read()` and `file.write()` used
`name.split('.')` to detect the file extension, which raises `AttributeError` for
`Path` objects.

## Changes

| File | What changed |
|---|---|
| `wrangles/recipe.py` | Added `os.PathLike` → `str` conversion at the top of `_load_recipe()` |
| `wrangles/connectors/file.py` | Same conversion at the top of `read()` and `write()` |
| `tests/recipes/test_run.py` | Added `test_run_path_object`, `test_run_pure_posix_path` |
| `tests/connectors/test_file.py` | Added `TestPathObjects` with 4 read/write tests |

The fix is a minimal, non-breaking guard — one `isinstance` check per entry point.
Using `os.PathLike` (rather than `pathlib.Path` directly) means it also covers
`PurePosixPath`, `PureWindowsPath`, and any third-party path-like type.

### `recipe.run()` with a Path

```python
import pathlib
import wrangles

recipe_path = pathlib.Path('recipes') / 'clean_data.wrgl.yml'

# Before: had to call str() explicitly
df = wrangles.recipe.run(str(recipe_path))

# After: Path object accepted directly
df = wrangles.recipe.run(recipe_path)
```

### `file.read()` with a Path

```python
import pathlib
import wrangles

data_file = pathlib.Path('data') / 'raw' / 'input.csv'

df = wrangles.connectors.file.read(data_file)
```

### `file.write()` with a Path

```python
import pathlib
import wrangles

output_dir = pathlib.Path('data') / 'processed'
output_dir.mkdir(parents=True, exist_ok=True)

wrangles.connectors.file.write(df, output_dir / 'output.xlsx')
```

### Typical real-world pattern — recipe next to the script

```python
from pathlib import Path
import wrangles

# __file__ gives the location of the current script,
# so the recipe path is always relative to it — no matter where you run from.
HERE = Path(__file__).parent
recipe = HERE / 'recipes' / 'clean_parts.wrgl.yml'

df = wrangles.recipe.run(recipe)
```

### Inside a recipe via the file connector

```yaml
read:
  - file:
      name: data/input.csv   # string paths still work as before

wrangles:
  - convert.case:
      input: Find
      output: Find
      case: upper

write:
  - file:
      name: data/output.csv
```

### DIFF
--- a/tests/connectors/test_file.py
+++ b/tests/connectors/test_file.py
@@ -1,6 +1,7 @@
 """
 Test the file connector for reading and writing files to the local file system.
 """
+import pathlib
 import uuid as _uuid
 import wrangles
 import pandas as _pd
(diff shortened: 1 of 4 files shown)


thomasstvr:
I'm not sure if this is what @ebhills had in mind. To me, the issue reads as if this would be an extra parameter of some sort.

ebhills:
the PR seems to be correct.  @thomasstvr -- the problem this alleviates is having to convert path obejcts to strings versus just passing them as is.  I can walk you through this example if helpful.

```python
def p2s(path: Path) -> str:
    """
    Convert a Path object to a string. This is useful for passing file paths to libraries that expect strings.
    """
    return str(path)
"""

HERE = Path(__file__).resolve().parent
ROOT = HERE.parent.parent
CUSTOMER = HERE.parent.name   # "RSIS"

recipe = p2s(HERE / "generate_upload.recipe")
file = p2s(ROOT / "data" / CUSTOMER / "Attribute Matching" / "checkpoints" / "1_joined_data_Gaskets, O-Rings, and Seals.parquet")
```

with the change we don't need p2s.

ebhills (commented):
Need to update the file: name schema to explain can be str or path

Image

mborodii-prog:
@ebhills doc string is updtaed

ebhills (approved):

PR #989, opened 2026-05-19
Extract.ai renaming columns bug fixes

### LINKED ISSUE #983: [BUG] Extract.ai renaming columns
**Describe the bug**
When using parenthesis in the first item in an extract.ai wrangle, the name is intermittently changed (typically the last parenthesis is dropped or everything in parenthesis is dropped). This is likely similar to issue #711 where single outputs with a space in the name were being written without the space intermittently, which is how things were returned from openai.

**To Reproduce**
1. Run an extract.ai such as this on a hand full of data.
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/15122/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/15122/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

<style>

</style>
</head>

<body link="#467886" vlink="#96607D">

Find | Description | Type | Default | Examples
-- | -- | -- | -- | --
Size (Diameter) | The diameter of the cap measured   in inches using astm standards | string |   | 1-7/8", 2-3/8",   2-7/8"
Size | The size   of the product | string |   | 1-7/8x2-3/8"

</body>

</html>

**Expected**
Output names should not be changed.

mborodii-prog:
## Problem

When an `extract.ai` output schema uses column names with parentheses, spaces, or other
special characters (e.g. `"Size (Diameter)"`), the OpenAI model occasionally returns a
slightly mutated key — a trailing colon, a dropped parenthesis, a space substitution.
Because the returned key no longer exactly matches the schema key, the recipe silently
writes an empty string to that column with no warning or error.

## Root cause

`wrangles/extract.py` passes the output dict keys verbatim as JSON-schema property names.
OpenAI's structured output mode works best with `[a-zA-Z0-9_]` identifiers — names with
parentheses or spaces are valid JSON but the model occasionally mutates them when echoing
them back, producing a key mismatch downstream.

The recipe layer (`recipe_wrangles/extract.py:261`) fills any unrecognised column with
`""` rather than raising, making the failure invisible.

---

## Fix

**File:** `wrangles/extract.py` — `ai()` function

Two-step sanitize-and-remap, applied only when output keys contain special characters
(zero overhead for clean keys):

1. **Before the API call** — replace any character outside `[a-zA-Z0-9_]` with `_`,
   deduplicate with a numeric suffix if needed, and build a reverse mapping.

2. **After the API call** — walk the results list and rename every key back to its
   original using the reverse mapping.

```python
# sanitize  (before API call)
"Size (Diameter)"  →  "Size__Diameter_"
"Temperature (F)"  →  "Temperature__F_"
"Weight (lbs)"     →  "Weight__lbs_"

# remap  (after API call)
"Size__Diameter_"  →  "Size (Diameter)"   ✓
"Temperature__F_"  →  "Temperature (F)"   ✓
"Weight__lbs_"     →  "Weight (lbs)"      ✓
```

The sanitized names are only ever seen by the model — the dataframe always receives the
original column names the user specified.

**Collision handling:** if two original keys would map to the same sanitized form, a
numeric suffix is appended (`_2`, `_3`, …) so the remap is always bijective.

---

Tested against the real OpenAI API (`gpt-4o-mini`) using three scenarios:

| Scenario | Schema width | Calls made | Empty-column events (before fix) | Empty-column events (after fix) |
|----------|-------------|-----------|-----------------------------------|---------------------------------|
| 1 — 2-key schema (issue example) | 2 keys | 400 row-calls | < 0.25 % (intermittent) | **0 / 400** |
| 2 — 8-key wide schema (raw API) | 8 keys | 300 row-calls | ~6–7 % of runs | **0 / 300** |
| 3 — 8-key wide schema (recipe) | 8 keys | 60 runs × 5 rows | silent empties on ~1 in 10 runs | **0 / 60 runs** |

Scenario 2 was the most reliable reproducer pre-fix: with 8 unit-parenthesis keys the
model produced a renamed key roughly once every 15 calls. After the fix, zero renames
were observed across 300 + 300 + 300 row-calls.

---

### DIFF
--- a/tests/recipes/wrangles/test_extract.py
+++ b/tests/recipes/wrangles/test_extract.py
@@ -1,6 +1,7 @@
 import pytest
 import wrangles
 import pandas as pd
+from unittest.mock import patch
 
 
 class TestExtractAddress:
@@ -4046,3 +4047,38 @@ class TestExtractAI:
             })
         )
         assert df['numbers'][0] >= 7
+
+    def test_ai_special_char_column_names(self):
+        """
+        Regression test for #983: column names containing parentheses or other
+        special characters must be preserved in the output dataframe.
+        The model receives sanitized property names; the recipe must remap them
+        back to the originals.
+        """
+        # Return values keyed by the sanitized names that were sent to the model
+        mock_responses = [
+            {"Size__Diameter_": '1-7/8"', "Size": '1-7/8x2-3/8"'},
+            {"Size__Diameter_": '2-3/8"', "Size": ""},
+        ]
+        with patch("wrangles.openai.chatGPT", side_effect=mock_responses):
+            df = wrangles.recipe.run(
+                """
+                wrangles:
+                - extract.ai:
+                    input: Product
+                    model: gpt-4o-mini
+                    api_key: dummy
+                    output:
+                      Size (Diameter):
+                        type: string
+                        description: The diameter of the cap in inches
+                      Size:
+                        type: string
+                        description: The overall size specification
+                """,
+                dataframe=pd.DataFrame({
+                    "Product": ['1-7/8" cap', '2-3/8" cap'],
+                }),
+            )
+        assert list(df.columns) == ["Product", "Size (Diameter)", "Size"]
+        assert df["Size (Diameter)"].tolist() == ['1-7/8"', '2-3/8"']
--- a/wrangles/extract.py
+++ b/wrangles/extract.py
@@ -248,6 +248,25 @@ def ai(
         for k, v in output.items()
     }
 
+    # Sanitize output keys: replace any character outside [a-zA-Z0-9_] with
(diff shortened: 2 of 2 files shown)


ebhills:
@mborodii-prog - wow - good catch! We should make this name sanitizer part of the core code so that we do it universally. Perhaps as part of wrangles 2.0?

thomasstvr (approved):

PR #993, opened 2026-05-20
992-refactor-lookup-actions-bug-request-key-colm

### LINKED ISSUE #992: [REFACTOR] Lookup Actions & [BUG] Request Key Colm
- [ ] When you run **upsert**, any columns not specified get deleted.  this is a BIG problem.  the behavior of train actions (other than overwrite) should be to maintain any data not explicitly requested to change (i.e. non-destructive). For example:
```yaml
  - train.lookup:
      model_id: ecca50a5-7fa0-404a # Category Attributes
      action: upsert
      columns:
        - Key
        - Schema # <- should not have to include - it exists and I do not want to update
        - Mandatory/SEQ # <- should not have to include - it exists and I do not want to update
        - Mapping
 ```
For my use case, I only need to update the Mapping column. But if that's all I specify then the other 2 get deleted.

- [ ] Please also review behavior when columns are not specified. **I think we should make it a mandatory parameter** and only add / update what is specified.

- [ ] Check other actions to ensure they also take this non-destructive approach to rows and columns
- [ ] Key should be able to be specified as both the only return value or one of multiple.  
  
  1. Currently if you just specify Key, it treats it like an arbitray name for the dict with all columns (except key!)
  2. And if you include it with another colm in the output you get an error!

Image

### PR DESCRIPTION
**Pull request by mborodii-prog**
## Summary

Two bugs in `train.lookup` write actions and the `lookup` wrangle are resolved.

---

### Problem

When running any non-overwrite action with a DataFrame that only contains a *subset* of the model's columns, the columns absent from the DataFrame were silently deleted from the model.

```python
# Model has: Key, Schema, Mapping
# User only wants to update Mapping for one row — Schema is intentionally omitted

wrangles.recipe.run("""
write:
  - train.lookup:
      model_id: <model_id>
      action: upsert
      variant: key
""", dataframe=pd.DataFrame({'Key': ['apple'], 'Mapping': ['green']}))

# BEFORE (bug): model now has Key, Mapping only — Schema was dropped
# AFTER (fix):  model still has Key, Schema, Mapping — Schema preserved
```

The same issue existed for `insert` (new rows had `NaN` in unspecified columns, which broke JSON serialisation) and `update`.

### Root cause

**`wrangles/connectors/train.py` — UPSERT branch (~line 345):**

```python
# BEFORE — slices the existing data to only the incoming columns
existing_df = existing_df_all[requested_cols]   # ← BUG
merged_df   = existing_df.copy()                # merge base is now partial
```

```python
# AFTER — uses ALL existing columns as the merge base
merged_df = existing_df_all.copy()
# New columns introduced by the upsert are added with '' for existing rows
new_cols = [c for c in requested_cols if c not in merged_df.columns]
for col in new_cols:
    merged_df[col] = ''
```

The same `existing_df` reference was also used in the non-key no-`MatchingColumns` path (`_pd.concat([existing_df, df])`), which is now `existing_df_all`.

For INSERT, `_pd.concat` can produce `NaN` for columns absent from new rows. Added `merged_df = merged_df.fillna('')` before building the JSON payload.

> `upsert` is the only action that accepts new columns — it treats them as an additive schema extension.  
> `insert` and `update` still raise `ValueError` when a column that doesn't exist in the model is supplied.

---

### Problem

```yaml
wrangles:
  - lookup:
      input: fruit
      output: Key           # wanted the key string back
      model_id: <model_id>
```

**Case A — `output: Key` only**  
`Key` is not in `metadata["settings"]["columns"]` (which lists only value columns).  
The routing fell into the "no named columns" branch, treating `Key` as an arbitrary output name and returning a full dict of all columns instead of the key string.

**Case B — `output: [Key, Schema]`**  
`Key` not in settings cols, `Schema` is → routing hit the "mixture" branch → `ValueError: Lookup may only contain all named or unnamed columns.`

### Root cause

**`wrangles/recipe_wrangles/main.py` — `lookup()` (~line 980):**

```python
# routing check used wrangle_output directly, which included 'Key'
if all([col in metadata["settings"]["columns"] for col in wrangle_output]):
    ...
elif not any([col in metadata["settings"]["columns"] for col in wrangle_output]):
    ...
else:
    raise ValueError('Lookup may only contain all named or unnamed columns.')
```

### Fix

Before the routing check, `Key` entries are extracted from `wrangle_output`. The routing and API call use only the remaining non-Key columns. After the lookup completes, any requested `Key` output column is populated directly from the input values.

```python
# Strip 'Key' from routing — it echoes the input, not a model value column
_key_indices  = {i for i, col in enumerate(wrangle_output) if col == 'Key'}
_key_out_cols = [output[i] for i in sorted(_key_indices)]
_wrangle_cols = [col for i, col in enumerate(wrangle_output) if i not in _key_indices]
_out_cols     = [col for i, col in enumerate(output)         if i not in _key_indices]

# ... (routing + API call using _wrangle_cols / _out_cols) ...

# Populate Key output column(s) with the input key values
for _key_col in _key_out_cols:
    df[_key_col] = df[input].values
```

Applied consistently across `by_row`, `by_dataframe`, and `by_matrix` modes.

### Examples after fix

```python
import wrangles, pandas as pd

df = pd.DataFrame({'fruit': ['apple', 'banana', 'cherry']})

# Case A — Key only
result = wrangles.recipe.run("""
wrangles:
  - lookup:
      input: fruit
      output: Key
      model_id: <model_id>
""", dataframe=df)
# result['Key'] → ['apple', 'banana', 'cherry']  ✓

# Case B — Key + value column together
result = wrangles.recipe.run("""
wrangles:
  - lookup:
      input: fruit
      output:
        - Key
        - Schema
      model_id: <model_id>
""", dataframe=df)
# result['Key']    → ['apple', 'banana', 'cherry']  ✓
# result['Schema'] → ['fruit', 'fruit',  'fruit']   ✓
```

---

### DIFF
--- a/wrangles/connectors/train.py
+++ b/wrangles/connectors/train.py
@@ -234,16 +234,17 @@ class lookup():
             description: Specific model to read
         """
 
-    def write(df: _pd.DataFrame, name: str = None, model_id: str = None, settings: dict = None, variant: str = 'key', action: str = 'overwrite') -> None:
+    def write(df: _pd.DataFrame, columns: list = None, name: str = None, model_id: str = None, settings: dict = None, variant: str = 'key', action: str = 'overwrite') -> None:
         """
         Train a new or existing lookup wrangle
 
         :param df: DataFrame to be written to a file
+        :param columns: Columns to include. Required for INSERT, UPDATE and UPSERT — only the listed columns will be added or updated.
         :param name: Name to give to a new Wrangle that will be created
         :param model_id: Model to be updated. Either this or name must be provided
         :param settings: Specific settings to apply to the wrangle
         :param variant: Variant of the Lookup Wrangle that will be created (key or semantic)
-        :param action: Action to take when training the lookup wrangle (insert, update, upsert)
+        :param action: Action to take when training the lookup wrangle (overwrite, insert, update, upsert)
         """
         _logging.info(f": Training Lookup Wrangle")
         if settings is None:
@@ -256,6 +257,18 @@ class lookup():
         # Normalize inputs and avoid mutating default args
         act = (action or 'overwrite').upper()
         settings = dict(settings or {})
+
+        # columns is required for partial-update actions
+        if columns is None and act in ('INSERT', 'UPDATE', 'UPSERT'):
+            raise ValueError(
+                f"Lookup: 'columns' is required for action '{action}'. "
+                "Specify the columns to add or update so that unrelated columns are not modified."
+            )
+
+        # Filter the DataFrame to only the requested columns
+        if columns is not None:
+            columns = _wildcard_expansion(df.columns, columns)
+            df = df[columns]
 
         def _get_columns_from_payload(payload):
             # payload can be a DataFrame or the 'tight' dict produced by _to_tight
@@ -333,18 +346,8 @@ class lookup():
                     columns=existing_content['Columns']
                 )
 
-                # Validate column compatibility with existing model
-                requested_cols = new_data['Columns']
-                missing_in_existing = [c for c in requested_cols if c not in existing_df_all.columns]
-                if missing_in_existing:
-                    raise ValueError(
-                        "Lookup: The following columns are not present in the existing model: "
-                        + ", ".join(missing_in_existing)
-                    )
-
-                existing_df = existing_df_all[requested_cols]  # Ensure same column order
-
                 # For key variant, ensure new data contains Key column
+                requested_cols = new_data['Columns']
                 normalized_variant = settings.get('variant', variant)
                 if normalized_variant == 'key' and 'Key' not in df.columns:
                     raise ValueError("Lookup: 'Key' column must be provided for 'key' variant")
@@ -353,13 +356,17 @@ class lookup():
                 updated = 0
 
                 # Merge data - avoid duplicates based on Key column
-                if variant== 'key' and 'Key' in existing_df.columns and 'Key' in df.columns:
+                if variant == 'key' and 'Key' in existing_df_all.columns and 'Key' in df.columns:
                     if df['Key'].duplicated().any():
                         raise ValueError("Lookup: All Keys must be unique")
-                    existing_keys = set(existing_df['Key'].tolist())
+                    existing_keys = set(existing_df_all['Key'].tolist())
 
-                    # Start with current data
-                    merged_df = existing_df.copy()
+                    # Use ALL existing columns as the base so unspecified columns are preserved.
+                    # New columns introduced by this upsert are added with '' for existing rows.
+                    merged_df = existing_df_all.copy()
+                    new_cols = [c for c in requested_cols if c not in merged_df.columns]
+                    for col in new_cols:
+                        merged_df[col] = ''
 
                     # Apply updates for matching keys
                     for _, row in df.iterrows():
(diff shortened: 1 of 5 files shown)


mborodii-prog:
@thomasstvr @ebhills as we discussed yesterday I made columns parameter mandatory ## Breaking change: `columns` is now required for `insert`, `update`, and `upsert`

`train.lookup` (Python API and recipe) now requires an explicit `columns` list when using any partial-update action (`insert`, `update`, `upsert`). Calling one of these actions without `columns` raises a `ValueError`.

```
ValueError: Lookup: 'columns' is required for action 'upsert'.
Specify the columns to add or update so that unrelated columns are not modified.
```

`overwrite` is unaffected — `columns` remains optional there.

---

### Migration

**Before** (will now raise `ValueError`):

```python
# Python API
wrangles.connectors.train.lookup.write(
    df,
    model_id='<model-id>',
    action='upsert',
    variant='key',
)
```

```yaml
# Recipe
write:
  - train.lookup:
      model_id: <model-id>
      action: upsert
      variant: key
```

**After** (explicit `columns` required):

```python
# Python API
wrangles.connectors.train.lookup.write(
    df,
    model_id='<model-id>',
    columns=['Key', 'Mapping', 'Weight'],
    action='upsert',
    variant='key',
)
```

```yaml
# Recipe
write:
  - train.lookup:
      model_id: <model-id>
      action: upsert
      columns:
        - Key
        - Mapping
        - Weight
      variant: key
```

---

### Behaviour by action

| Action | `columns` required? | Effect |
|--------|-------------------|--------|
| `overwrite` | No | Replaces the entire model with the DataFrame |
| `insert` | **Yes** | Appends rows whose Key doesn't exist yet; only the listed columns are written |
| `update` | **Yes** | Updates only existing keys; only the listed columns change |
| `upsert` | **Yes** | Updates existing keys, inserts new ones; only the listed columns are touched |

---

### Wildcard support

`columns` accepts wildcard patterns via the existing `_wildcard_expansion` helper, so patterns like `*` or `Mapping*` work as expected:

```yaml
columns:
  - Key
  - Mapping*   # matches Mapping, MappingDetail, etc.
```

---

### Example: upsert preserving unspecified columns

Model state before upsert — columns: `Key`, `Schema`, `Mapping`, `Mandatory`

```python
update_df = pd.DataFrame({
    'Key':     ['apple'],
    'Mapping': ['green'],
    'Weight':  [1.0],    # new column not yet in the model
})

wrangles.connectors.train.lookup.write(
    update_df,
    model_id='<model-id>',
    columns=['Key', 'Mapping', 'Weight'],
    action='upsert',
    variant='key',
)
```

Model state after upsert:

| Key    | Schema | Mapping | Mandatory | Weight |
|--------|--------|---------|-----------|--------|
| apple  | fruit  | green   | true      | 1.0    |
| banana | fruit  | yellow  | false     |        |
| cherry | fruit  | red     | true      |        |

- `Schema` and `Mandatory` on the `apple` row are **preserved** (not in `columns`, not touched).
- `Weight` is a new column — existing rows get an empty string.

thomasstvr:
@mborodii-prog is there a series of commits that did not get pushed or am I missing something? I can't find anything about columns being required, even the tests are passing without columns (which none seem to have).

mborodii-prog:
@thomasstvr  It's because it will be a breaking change and on one of 1:1 call with Eric we decided to postpone that change

wrangles/connectors/train.py
ebhills: @mborodii-prog Is this docstring what generates the schema?  In any case, need to specify all 4 values for the action. And given that the default is destructive, let's make it required param.

mborodii-prog: @ebhills updated the doc strings

mborodii-prog:
@ebhills @thomasstvr you can use recipe 'test lookup 922' for testing in QA

ebhills:
Queue triage (2026-07-27)

- **Next action:** This is the canonical PR for the branch (duplicate #1038 is closed). Resolve the open thread, rerun the cancelled checks, and re-request one reviewer for v1.20.

PR #995, opened 2026-05-25
Extract ai responses 2026

### LINKED ISSUE #792: extract.ai examples - don't always want comman splitting
Currently the wrangle assumes that if example is a string that it's a user shorthand for several examples. This isn't always true, sometimes the example is meant to contain commas.  Here's a workaround:

```
## The Code Fix
We'll add an and not value.strip().endswith('\\\\') condition. This checks if the string, after trimming any whitespace, ends with a literal double backslash \\.

Python

# Parse any comma separated values into lists
node = {
    label: [x.strip() for x in value.split(",")]
    if (
        label in ("examples", "enum", "properties") and
        isinstance(value, str) and
        # ADDED CONDITION: Only split if the string does NOT end with \\
        not value.strip().endswith('\\\\')
    )
    else value
    for label, value in node.items()
}
```
How It Works
value.strip(): This removes any accidental leading or trailing whitespace from your example string.

.endswith('\\\\'): This is the key part. It checks if the string ends with \\. We need four backslashes in the code to represent a literal two-backslash string in Python, as the backslash is an escape character.

note: This inverts the logic, so the split(',') only runs when the string does not end with \\.

**FYI** fixed another bug where we incorrectly checked node['example'] instead of 'examples'.  This is already committed to main.

### LINKED ISSUE #1076: [BUG] extract.ai acts up when columns already exist
When an output column of extract.ai already exists, it affects the output of the extract. For example, by running the test below:

```python
df = wrangles.recipe.run(
    """
    wrangles:
    - create.column:
        output: length
        value: 123
        
    - extract.ai:
        model: gpt-4o-mini
        api_key: ${OPENAI_API_KEY}
        seed: 1
        timeout: 60
        retries: 2
        output:
            length:
            type: string
            description: >-
                Any lengths found in the data
                such as cm, m, ft, etc.
    """,
    dataframe=pd.DataFrame({
        "data": [
            "wrench 25mm",
            "6m cable",
            "screwdriver 3mm"
        ],
    })
)
```

The output becomes:

|  Index | data | length |
|:-------:|:-----:|:-------:|
| 0 | wrench 25mm | 123 |
| 1 | 6m cable | 6m |
| 2 | screwdriver 3mm | 123 |

Instead of the expected:

|  Index | data | length |
|:-------:|:-----:|:-------:|
| 0 | wrench 25mm | 25mm |
| 1 | 6m cable | 6m |
| 2 | screwdriver 3mm | 3mm |

mborodii-prog:
## Summary

Upgrades `extract.ai` to OpenAI's Responses API through a shared definition compiler and runtime used by YAML recipes, direct Python calls, and saved WranglesXL models. The implementation keeps an explicit legacy Chat Completions compatibility path while standardizing schemas, examples, defaults, caching, retries, and error handling across entry points.

### Unified definitions and schema compilation

- Adds a common compiler for recipe definitions and saved-model definitions.
- Makes output fields nullable by default while keeping every declared field present in the structured response.
- Supports `nullable: false` as an explicit opt-out.
- Normalizes both YAML `null` and the user-entered enum value `"null"` to JSON Schema nullability.
- Preserves flexible/dynamic objects: schemas with unknown properties use compatible provider settings and retain local validation.
- Accepts human-friendly pseudo-JSON examples through safe YAML parsing.
- Supports both field-level examples and holistic input/output examples.
- Maps saved-model `Examples` to `Example - Output` for backward compatibility and adds support for `Example - Input`.

### Responses API runtime

- Uses the Responses API by default with structured outputs.
- Retains Chat Completions through an explicit `protocol: chat_completions` compatibility mode.
- Temporarily infers legacy protocol from a legacy endpoint when needed and emits a warning so definitions can be upgraded.
- Maps compatible legacy parameters, rejects conflicting parameters, and warns when an option cannot be honored.
- Handles structured results, refusals, incomplete responses, rate limits, usage reporting, retries, and deadline exhaustion consistently.
- Sends reasoning and text-verbosity options only when supported by the selected model.

### Latency, concurrency, and cost controls

The packaged `extract_fast` defaults are:

| Setting | Default |
| --- | --- |
| Provider / protocol | `openai` / `responses` |
| Model | `gpt-5.4-mini` |
| Maximum concurrency | 32 |
| Per-request timeout | 12 seconds |
| Shared call deadline | 15 seconds |
| Retries | 1, bounded by the shared deadline |
| Reasoning effort | `none` when supported |
| Text verbosity | `low` when supported |
| Structured output | strict where the schema permits it |
| OpenAI response storage | disabled |

Each unique uncached row is still one synchronous OpenAI request; this does not use OpenAI's asynchronous Batch API or combine multiple dataframe rows into one model request. WranglesXL's recipe batch size remains the number of rows sent to the Lambda invocation, while the extract runtime schedules those rows concurrently within the same invocation.

### Two complementary cache layers

- Adds a bounded, warm-process result cache for complete extraction results: 1-hour TTL, 512 entries, 64 KiB maximum cached value, request de-duplication, and single-flight protection.
- Cache keys include the provider, protocol, credential identity hash, model, prompt, schema, options, endpoint, and row input.
- Errors, timeouts, and oversized values are not cached.
- Keeps the prompt's static prefix stable so OpenAI's automatic prompt caching can independently reduce repeated-prefix latency and input cost.
- Provides environment-variable kill switches and tuning controls for the local result cache.

### Documentation and examples

- Adds an operational README covering the XL-to-Lambda batch boundary, extract scheduling, deadlines, retries, both cache layers, and tuning guidance.
- Adds configuration documentation for Python, YAML, and saved-model definitions.
- Adds an end-to-end YAML classification-judge example and notebook examples.
- Expands compiler, cache, Responses API, recipe, compatibility, nullable-output, and example-handling test coverage.

## Compatibility notes

Existing definitions can be upgraded incrementally. Compatibility mappings are intentionally transitional and warn or fail clearly when a legacy option is ambiguous. Broader model-configuration work remains separate from this PR.

Python receives nullable values as `None`. The JSON/Excel presentation boundary remains responsible for rendering those values as the appropriate empty cell/value; the extraction layer does not replace `None` with an empty string.

## Additional maintenance

Includes narrow cleanup for warnings encountered while exercising the expanded test suite, including raw regular-expression strings, pandas frequency aliases, dataframe-copy handling, and a targeted filter for an upstream Apprise deprecation warning. It does not otherwise change Apprise notification behavior.

## Validation

The current PR head passes the **Push to Main** GitHub Actions workflow. Coverage includes the shared compiler, Responses request/response handling, compatibility behavior, retries/deadlines, caching and single-flight behavior, nullable schemas, saved-model mappings, and YAML recipe execution.

## Follow-ups outside this PR

- Add an explicit execution channel/profile so WranglesXL and unconstrained Python runs can select different latency policies without caller inference.
- Extend the provider abstraction to additional model providers.
- Redesign the WranglesXL saved-model interface for richer holistic-example authoring.

## Reference

[Original classification notebook](https://github.com/user-attachments/files/28342015/classify_with_wrangles_extract_ai.1.ipynb)


mborodii-prog:
new demo file: 
[demo_classify_complete.ipynb](https://github.com/user-attachments/files/28760826/demo_classify_complete.ipynb)

ebhills:
Queue triage (2026-07-27)

- **Primary reviewer:** @ebhills
- **Next action:** Review the focused behavior and test boundary, then approve or request only blocking changes.

ebhills (approved):

ebhills:
@codex - Please update this PR's summary description to reflect the most recent round of commits.

ebhills (approved):

PR #996, opened 2026-05-26
766 select.dictionary_element: add split parameter

### LINKED ISSUE #766: select.dictionary_element behavior when no output specified
I realize that the global wrangles default behavior is to overwrite columns when no output is specified.  However, I think select.dictionary_element against a single element should be different.

When calling it for a specific element, the wrangler clearly wants to 'extract' the info rather than write over the input dictionary. I have lost a lot of time troubleshooting why my dictionary is no longer a dict.

As is, the user is effectively forced to name it what it's already named.

mborodii-prog:
# select.dictionary_element: add split parameter

## Summary

Per @thomasstvr's suggestion, instead of adding a new `select.dict_element`
wrangle, this adds a `split` parameter to the existing
`select.dictionary_element` wrangle. Default behavior is unchanged.

## Motivation

`select.dictionary_element` overwrites the input column when no `output`
is specified, which is convenient for quick in-place replacements but
makes it easy to lose the original data (see #766). Rather than
introducing a separate wrangle with different default behavior, `split`
opts into the non-destructive behavior on the existing wrangle.

## Behavior

| `split` | Behavior |
|---|---|
| `false` (default) | Unchanged: writes to `output` if given, otherwise overwrites the input column |
| `true` | Writes the selected element(s) to new column(s) named after the key(s), preserving the input column. Cannot be combined with `output`. With a single key, creates one column named after that key; a list, wildcard, or regex element creates one column per matched key. |

## Examples

**Simple key** — new column named after the element, input preserved:
```yaml
wrangles:
- select.dictionary_element:
    input: product
    element: colour
    split: true
# input:  product = {'colour': 'red', 'shape': 'circle'}
# result: product = {'colour': 'red', 'shape': 'circle'}  (unchanged)
#         colour  = 'red'                                  (new)
```

**List of keys** — one column per key:
```yaml
wrangles:
- select.dictionary_element:
    input: properties
    element:
    - A
    - B
    split: true
# input:  properties = {'A': '1', 'B': '2', 'C': '3'}  (unchanged)
# result: A = '1', B = '2'                               (new columns)
```

**Wildcard** — matched keys resolved at runtime, each becomes a column:
```yaml
wrangles:
- select.dictionary_element:
    input: specs
    element: A*
    split: true
# input:  specs = {'A1': '1', 'B1': '2', 'A2': '3'}  (unchanged)
# result: A1 = '1', A2 = '3'                           (new columns)
```

**Regex** — same as wildcard:
```yaml
wrangles:
- select.dictionary_element:
    input: specs
    element: 'regex: B.*'
    split: true
# input:  specs = {'A1': '1', 'B1': '2', 'A2': '3'}  (unchanged)
# result: B1 = '2'                                     (new column)
```

### DIFF
--- a/tests/recipes/wrangles/test_main.py
+++ b/tests/recipes/wrangles/test_main.py
@@ -3469,7 +3469,7 @@ class TestTranslate:
                 target_language: EN-GB
         """
         df = wrangles.recipe.run(recipe, dataframe=data)
-        assert df.iloc[0]['English'] == 'Hello World!'
+        assert df.iloc[0]['English'] == 'Hello, world!'
 
     def test_translate_2(self):
         """
@@ -3487,7 +3487,7 @@ class TestTranslate:
                 target_language: English
         """
         df = wrangles.recipe.run(recipe, dataframe=data)
-        assert df.iloc[0]['English'] == 'Hello World!'
+        assert df.iloc[0]['English'] == 'Hello, world!'
 
     def test_translate_3(self):
         """
--- a/tests/recipes/wrangles/test_select.py
+++ b/tests/recipes/wrangles/test_select.py
@@ -377,6 +377,184 @@ class TestSelectDictionaryElement:
         )
         assert list(df.columns) == ['column', 'output column'] and len(df) == 0
 
+    def test_dictionary_element_no_output_multi_input_overwrites(self):
+        """
+        Without output, multiple input columns fall back to overwriting
+        each input column individually.
+        """
+        df = wrangles.recipe.run(
+            """
+            wrangles:
+            - select.dictionary_element:
+                input:
+                - Col1
+                - Col2
+                element: x
+            """,
+            dataframe=pd.DataFrame({
+                'Col1': [{'x': 1, 'y': 2}],
+                'Col2': [{'x': 10, 'y': 20}],
+            })
+        )
+        assert df['Col1'][0] == 1 and df['Col2'][0] == 10
+
+
+class TestSelectDictionaryElementSplit:
+    """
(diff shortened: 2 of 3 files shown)


thomasstvr:
Automatically outputting a split dictionary like this will be a breaking change. I would suggest leaving the default behavior as is and adding some sort of parameter that splits the selected elements.

mborodii-prog:
@thomasstvr it will not be a breaking change as it's new wrangle

thomasstvr:
@mborodii-prog somehow I completely missed that. In general, I don't think it is a good idea to just create a new wrangle when we don't like the behavior of another. I skipped right over it because the names are so similar myself.

mborodii-prog:
@thomasstvr I updated current PR approach that you proposed: added new parameter split instead of creating new wrangle

ebhills:
@thomasstvr - I don't think 'split' make senses in the context of select, especially since we have a split.dictionry.

While we generally don't want a breaking change, in this case I believe it is worthwhile.  I think I ran a models table query that found all of the instances where this wrangle is used without an output.  I will take the action item to rerun that and assess how we can change the behavior to be intuitive -> you specify the key(s) you want and they pop out as columns without having to specify output. That way a) it always yields columns, and b) the user only needs to specify output(s) when they want to rename.

I have never had a use for the current default behavior of 'selecting' a subset of keys but leaving it as a smaller dictionary.  Have you?

ebhills:
Queue triage (2026-07-27)

- **Primary reviewer:** @thomasstvr
- **Next action:** Review the split-parameter behavior and focused tests; approve or request blocking changes.

PR #997, opened 2026-05-26
add train.delete to delete a model by model_id

### LINKED ISSUE #991: [ENHANCEMENT] delete_wrangle wrangle
**Feature description**
Create a wrangle out of _delete_model. Let's discuss whether this should only be a python level wrangle or also a recipe wrangle. I don't see a need for it to be a recipe wrangle, but would like to know what others think.

**Use case**
Would be nice to be able to add tests that train new models, then use this to delete them.

**Acceptance criteria**
- [ ] It deletes wrangles by model id
- [ ] Tests are added

mborodii-prog:
Must be merged after API-Core changes https://github.com/wrangleworks/API-Core/pull/111

## Summary

- Adds `wrangles.train.delete(model_id)` as a static method on the `train` class in `wrangles/train.py`
- Accepts only `model_id` — no wrangle type required
- Raises `RuntimeError` if the API returns a non-OK response
- Not placed in `recipe_wrangles/main.py`, so it is never auto-discovered as a recipe wrangle or bound as a DataFrame accessor

### DIFF
--- a/tests/connectors/test_train.py
+++ b/tests/connectors/test_train.py
@@ -1679,9 +1679,15 @@ class TestTrainDelete:
     Tests for wrangles.train.delete
     """
 
-    def test_create_and_delete_model(self):
+    def _get_model_id(self, response):
+        body = response.json()
+        model_id = body.get('model_id') or body.get('id') or body.get('modelId') or body.get('model')
+        assert model_id, f"No model_id in creation response: {body}"
+        return model_id
+
+    def test_create_and_delete_classify(self):
         """
-        Train a new classify model then delete it by the returned model_id.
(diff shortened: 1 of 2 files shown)


thomasstvr (changes requested):
We are going to need more tests, at least one for each type of wrangle the train connector can create. At the same time, (if I remember correctly) there aren't very many tests that train new models. Can you make sure each train connector is properly tested for creating a new model and add this delete to each one? That would take care of both cases with one test.

ebhills:
Queue triage (2026-07-27)

- **Next action:** Add the requested delete-path coverage and answer the requested changes, then re-request one primary reviewer. Keep the public API change focused here.

PR #1007, opened 2026-06-09
Fix where skipping wrangles when all rows are filtered ou

### LINKED ISSUE #1005: [BUG] Where passes empty dataframe
**Describe the bug**
When where filters out the entire batch/dataframe, and empty dataframe is still passed. This causes errors if trying to to use where to only run wrangles when a column exists.

**To Reproduce**
```yaml
wrangles:
  - recipe:
      where: successful_search == True # All false in this case
      wrangles:
        - split.dictionary:
            input: scored_results # Exists, but is an empty dict. Expected to contain summary key

        - split.dictionary:
            input: summary # Does not exist due to the above and causes an error
```

In the example above, successful_search is simply a marker that can be used to tell if scored_results is a dictionary with the expected keys. 

Works:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/15122/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/15122/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

scored_results | successful_search
-- | --
{'summary': 'this is a   summary', 'other key': 'this is some other key'} | TRUE
{} | FALSE

</body>

</html>

Doesn't work:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/15122/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/15122/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

scored_results | successful_search
-- | --
{} | FALSE
{} | FALSE

</body>

</html>

**Expected**
It should skip passing empty dataframes to wrangles, even though wrangles are also expected to work on empty dataframes.

**Acceptance criteria**
- [ ] Works without slowing things down or causing other issues
- [ ] Well tested

mborodii-prog:
### Problem

When a `where` clause matched **no rows**, the resulting empty DataFrame was still passed into the wrangle for execution. This became an issue for chained wrangles where an earlier step creates a column that a later step depends on — because on an empty DataFrame no new columns are ever produced.

**Failing example:**

```yaml
wrangles:
  - recipe:
      where: successful_search == True   # all False → empty DataFrame
      wrangles:
        - split.dictionary:
            input: scored_results        # runs but creates no columns (empty df)
        - split.dictionary:
            input: summary               # KeyError: column 'summary' does not exist
```

With input data:

| scored_results | successful_search |
|---|---|
| `{}` | `False` |

**Error produced:**

```
KeyError: 'ERROR IN WRANGLE #1 recipe - "ERROR IN WRANGLE #2 split.dictionary - \'Column summary does not exist\'"'
```

### Fix

In `wrangles/recipe.py`, after `_filter_dataframe()`, added a four-line guard — if `where` produces an empty DataFrame, restore the original and `continue` to the next wrangle without executing:

```python
df = _filter_dataframe(df, where=params.get('where'), ...)

# If where filters out all rows, skip the wrangle entirely
if len(df) == 0:
    df = df_original
    continue
```

### Behaviour after fix

**All rows filtered out → wrangle skipped, data unchanged:**

| scored_results | successful_search |
|---|---|
| `{}` | `False` |

No error, no spurious columns added.

**Partial match → still works correctly:**

| scored_results | successful_search | summary | other key |
|---|---|---|---|
| `{'summary': 'this is a summary', 'other key': '...'}` | `True` | `this is a summary` | `...` |
| `{}` | `False` | `` | `` |

Only the matching row is processed; unmatched rows are merged back unchanged.


ebhills (changes requested):
There a column shift bug. I am running this recipe w batch size = 1:
```yaml
wrangles:
  - convert.case:
      input: Desc
      output: output_column_name
      case: upper
      where: WC = "A"
```

Image

mborodii-prog:
@ebhills now it must be working fine (batch_size = 1)
image

ebhills (approved):

PR #1008, opened 2026-06-09
Fix DataFrame accessor mutating original on assignment

### LINKED ISSUE #873: dataframe function writes to existing dataframe when trying to create a new df from the result
I am using the dataframe.py functions to run wrangles in jupyter nb.  I discovered that a wrangle like this actually modifies df:
```yaml
df_sample = df.wrangles.select.sample(rows=10)
```
See image below for full example.

I recall that Chris defaulted to this behavior so that we could run wrangles with simple calling syntax (no reassignment necessary).
```yaml
df.wrangles.select.sample(rows=10)
```

But ideally if the user wants the results to go into a new df (like in my example) that would work too. Perhaps the new Copy on Write behavior in pandas 3.0 will make this easier.

Image

mborodii-prog:
### Problem

When using the `.wrangles` accessor and assigning the result to a new variable, the **original** DataFrame was also mutated. Both variables ended up pointing to the same in-memory object.

```python
df = DataFrame({'col': range(20)})
df_sample = df.wrangles.select.sample(rows=5)

print(len(df))        # 5  ← should still be 20
print(len(df_sample)) # 5
print(df is df_sample) # True ← same object
```

The same problem affected any wrangle that modifies columns:

```python
df = DataFrame({'name': ['alice', 'bob', 'carol']})
df_upper = df.wrangles.convert.case(input='name', case='upper')

print(df['name'].tolist())       # ['ALICE', 'BOB', 'CAROL'] ← should be unchanged
print(df_upper['name'].tolist()) # ['ALICE', 'BOB', 'CAROL']
print(df is df_upper)            # True
```

### Root cause

Every `make_method` closure in `dataframe.py` used this pattern:

```python
def method(self, *args, **kwargs):
    self._df.__init__(target_func(self._df, *args, **kwargs))  # reinitialises df in-place
    return self._df                                             # returns the same object as df
```

Two things go wrong:

1. **`self._df.__init__(result)`** calls `__init__` on the existing object, which reinitialises its data in-place without changing its identity (`id()`). So `df` is mutated.
2. **`return self._df`** hands back the same reference as `df`. Any assignment (`df_sample = ...`) just creates a second name for the same object.

Many wrangle functions also assign columns directly (`df[col] = ...`), so even without the `__init__` trick the original would be modified.

The pattern appeared in four places: `_wrangles_accessor.make_method`, `_wrangles.make_method`, `_read.make_method`, and `_read.file`.

### Fix

Pass a copy of the DataFrame into `target_func` and return its result directly — no `__init__`, no returning `self._df`:

```python
# Before
def method(self, *args, **kwargs):
    self._df.__init__(target_func(self._df, *args, **kwargs))
    return self._df

# After
def method(self, *args, **kwargs):
    return target_func(self._df.copy(), *args, **kwargs)
```

Using `.copy()` means wrangle functions that assign columns in-place operate on a throwaway copy, so the caller's original DataFrame is never touched. The returned value is always an independent object.

### Behaviour after fix

**Assigning to a new variable — original unchanged:**

```python
df = DataFrame({'col': range(20)})
df_sample = df.wrangles.select.sample(rows=5)

print(len(df))         # 20 ✓
print(len(df_sample))  # 5  ✓
print(df is df_sample) # False ✓
```

```python
df = DataFrame({'name': ['alice', 'bob', 'carol']})
df_upper = df.wrangles.convert.case(input='name', case='upper')

print(df['name'].tolist())       # ['alice', 'bob', 'carol'] ✓
print(df_upper['name'].tolist()) # ['ALICE', 'BOB', 'CAROL'] ✓
print(df is df_upper)            # False ✓
```

**Reassignment still works as before:**

```python
df = DataFrame({'col': ['hello', 'world']})
df = df.wrangles.convert.case(input='col', case='upper')

print(df['col'].tolist())  # ['HELLO', 'WORLD'] ✓
```

**Multiple calls on the same DataFrame — each independent:**

```python
df = DataFrame({'col': ['hello', 'world', 'foo', 'bar', 'baz']})
_ = df.wrangles.convert.case(input='col', case='upper')
_ = df.wrangles.select.sample(rows=2)

print(df['col'].tolist())  # ['hello', 'world', 'foo', 'bar', 'baz'] ✓
print(len(df))             # 5 ✓
```

### DIFF
--- a/tests/test_dataframe.py
+++ b/tests/test_dataframe.py
@@ -1,0 +1,49 @@
+import pandas as pd
+import pytest
+from wrangles.dataframe import DataFrame
+
+
+class TestWranglesAccessor:
+    """Tests for the .wrangles DataFrame accessor (issue #873)."""
+
+    def test_select_sample_does_not_mutate_original(self):
+        df = DataFrame({'col': range(20)})
+        df_sample = df.wrangles.select.sample(rows=5)
+        assert len(df) == 20
+        assert len(df_sample) == 5
+
+    def test_select_sample_returns_independent_object(self):
+        df = DataFrame({'col': range(20)})
+        df_sample = df.wrangles.select.sample(rows=5)
+        assert df is not df_sample
+
+    def test_convert_case_does_not_mutate_original(self):
+        df = DataFrame({'name': ['alice', 'bob', 'carol']})
+        df_upper = df.wrangles.convert.case(input='name', case='upper')
+        assert df['name'].tolist() == ['alice', 'bob', 'carol']
+        assert df_upper['name'].tolist() == ['ALICE', 'BOB', 'CAROL']
+
+    def test_convert_case_returns_independent_object(self):
+        df = DataFrame({'name': ['alice', 'bob']})
+        df_upper = df.wrangles.convert.case(input='name', case='upper')
+        assert df is not df_upper
+
+    def test_split_dictionary_does_not_mutate_original(self):
+        df = DataFrame({'data': [{'a': 1, 'b': 2}]})
+        df_split = df.wrangles.split.dictionary(input='data')
+        assert 'a' not in df.columns
+        assert 'a' in df_split.columns
+
+    def test_reassignment_updates_variable(self):
+        """Reassigning df = df.wrangles... should still work correctly."""
+        df = DataFrame({'col': ['hello', 'world']})
+        df = df.wrangles.convert.case(input='col', case='upper')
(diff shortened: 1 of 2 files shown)


ebhills:
Queue triage (2026-07-27)

- **Primary reviewer:** @thomasstvr
- **Next action:** Review the DataFrame copy/identity behavior and regression tests.

ebhills (approved):

PR #1014, opened 2026-06-17
1013 update issue templates

### LINKED ISSUE #1013: Update issue templates
1. Do not put [BUG] etc in the title.  Also, do not set that as the Label.  This value should be set as the Type.

2. Let's limit the 'form' to 3 required text blocks:
  i. Description
  ii. Desired Behavior
 iii. Example current / desired

3. Auto-assign you? Auto-assign priority (P1 for enhancement, P0 for Bug).  Maybe even auto-assign the Label using CoPilot (since that assignment would be 'semantic')

mborodii-prog:
Simplify issue templates and add auto-assign workflow (#1013)

  - Remove [TYPE] prefixes from all issue template titles
  - Replace type-based labels with priority labels (P0 for Bug, P1 for all others)
  - Add `type:` frontmatter field to all templates
  - Slim all 8 templates to 3 consistent sections: Description, Desired Behavior, Example current/desired
  - Add issue-management workflow to auto-assign new issues to mborodii-prog

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,31 +1,17 @@
 ---
 name: Bug
 about: An unexpected problem or behavior
-title: '[BUG] '
-labels: bug
+title: ''
+labels: 'P0'
 assignees: ''
+type: Bug
 ---
 
-**Describe the bug**
+**Description**
 What happened?
 
-**To Reproduce**
-1. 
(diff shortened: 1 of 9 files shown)


ebhills (changes requested):
Let's walk through together:
1. Want to use new GH feature: issue Fields
2. Add checklist of features (where, default, new params, expansion, ...)

PR #1018, opened 2026-06-19
1016 refactor hardcoded OpenAI models

### LINKED ISSUE #1016: [REFACTOR] Hardcoded OpenAI Models
**What needs refactoring**
calls to specific OpenAI models in our code and tests

**Current state**
hard coded to older models.  Some will probably break soon b/c of deprecation. Also, creates lots of noise in our usage and cost data.

**Proposed changes**
Make models a top-level, global CONFIG VAR. Point all AI functions and tests at this VAR.
Create specific keys for various functions (eg PY Testing, Schema Standardization, ...) so that we have granular visibility of usage / costs.

**Benefits**
- Performance improvement
- Better maintainability
- Code quality
- Other:

**Breaking changes**
- [ ] Yes
- [ ] No

**Details**

**Acceptance criteria**
- [ ] 
- [ ] 
- [ ]

mborodii-prog:
## Summary

Removes hardcoded OpenAI model name strings scattered across the library and centralizes them in a single config location, so models can be updated in one place and production/test usage can be tracked separately in OpenAI's usage/cost data.

- Added `wrangles.config.models` with `extract_ai`, `generate_ai`, and `embeddings` defaults, plus a nested `models.testing` class holding the model names the test suite should use (kept distinct from production so test runs are attributable separately in OpenAI usage data).
- Replaced hardcoded model defaults with references to `config.models.*` in:
  - `wrangles/extract.py` (`extract.ai`)
  - `wrangles/generate.py` (`generate.ai`)
  - `wrangles/openai.py` (`embeddings`)
  - `wrangles/recipe_wrangles/create.py` (`create.embeddings`)
  - `wrangles/recipe_wrangles/generate.py` (`generate.ai` recipe wrangle)
- Updated tests (`test_extract.py`, `test_generate.py`, `test_create.py`, `test_wrangles.py`) to pull model names from `config.models.testing.*` instead of hardcoding real model identifiers. Recipe YAML strings use a `MODEL_PLACEHOLDER` token swapped via `.replace()` rather than an f-string, since the YAML already contains `${OPENAI_API_KEY}`-style placeholders that would otherwise collide with f-string interpolation.

## Why `.replace()` instead of f-strings in tests

Several recipe tests build YAML as a plain string containing `${OPENAI_API_KEY}`. An f-string would try to evaluate `{OPENAI_API_KEY}` as a Python expression and raise `NameError`. Using a `MODEL_PLACEHOLDER` token with `.replace()` avoids that collision while keeping no real/hardcoded model name in the test source.

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -1,0 +1,48 @@
+# WranglesPY development container
+
+The shared `devcontainer.json` is intentionally host-neutral. VS Code on Windows
+and GitHub Codespaces both use the same Python 3.13 container definition.
+
+## Windows with Docker Desktop
+
+Docker Desktop's Linux engine must be ready before VS Code evaluates the
+devcontainer. From the repository root, run:
+
+```powershell
+.\.devcontainer\wait-for-docker.ps1
+```
+
+The helper starts Docker Desktop when necessary and waits until the selected
+Docker context can reach a Linux server. When it reports that Docker is ready,
+open the repository in VS Code and run:
+
+```text
+Dev Containers: Reopen in Container
+```
+
+Use `Dev Containers: Rebuild Container` after changing `devcontainer.json` or
+the dependency files.
+
+`Dev Containers: Attach to Running Container...` remains useful as a recovery
+option when this repository's container is already running. It is not the
+normal workflow because attaching does not necessarily apply this repository's
+creation commands and VS Code customizations.
+
+## GitHub Codespaces
+
+Create or reopen the Codespace normally. Codespaces reads
+`.devcontainer/devcontainer.json` directly. The Windows readiness helper is not
+referenced by the configuration and does not run in Codespaces.
+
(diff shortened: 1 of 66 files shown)


ebhills (changes requested):
we need to set up the API keys and incorporate that into the build / test process.  Let's do it with Leo's new OIDC flow so that we don't have to store the keys here.

mborodii-prog:
@thomasstvr @ebhills updated code to use new API_KEY secrets for running tests

ebhills:
Queue triage (2026-07-27)

- **Next action:** Address the requested configuration/OIDC/test-boundary changes, keeping one configuration authority, then re-request one reviewer. Retain v1.20 only if those decisions are resolved before freeze.

ebhills:
@copilot+gpt-5.6-sol:medium resolve the merge conflicts in this pull request

PR #1038, opened 2026-07-01
992 refactor lookup actions bug request key colm

### LINKED ISSUE #992: [REFACTOR] Lookup Actions & [BUG] Request Key Colm

mborodii-prog:


ebhills:
Queue triage (2026-07-27): closing as a duplicate of #993. Both PRs use the same head branch; #993 is the canonical review conversation.

PR #1063, opened 2026-07-13
985 enhancement allow list of one/multiple defaults

mborodii-prog:
Check CI on dev after widening the `test_multiprocess` timing tolerance to fix Windows CI flakiness (process-spawn overhead was pushing wall-clock time past the old 10s upper bound).


mborodii-prog:
Closing — replacing with a PR built from just this branch's own commits, not everything pulled in from main.

PR #1069, opened 2026-07-14
delete model after creating

### LINKED ISSUE #1049: Tests Creating Large Number of Actual Wrangles
**Describe the bug**
reviewing the models table noted that we now have 36K wrangles. of those, 29K are created by wrwx@wrangle.works.  Looking at the 16K lookups for example, the name makes it clear that (all?) are being generated by tests

**Expected**
wrangles generated by tests should be deleted in a timely fashion if not immediately. And/or we should mock them

**Additional context**

mborodii-prog:
## Summary

  Integration tests for `train.classify`, `train.extract`, `train.lookup`, and `train.standardize` create real models on the WrangleWorks platform but never clean them up, leaving a growing number of orphaned test models in the account. This adds a `wrangles.train.delete()` method and updates the affected tests to delete the models they create.

  - Add `wrangles.train.delete(model_id)`, which calls the `DELETE /model/delete` API endpoint and raises `RuntimeError` on a non-OK response.
  - Wrap model-creation tests in `try/finally` so the created model is deleted after the test runs, regardless of pass/fail.
  - Add `_delete_model_from_log()` helper to `tests/connectors/test_train.py` to extract a model ID from a logged message (e.g. `"New classify model created :: <id>"`) and delete it, tolerating failure while cleaning up.
  - Add `TestTrainDelete` test class covering create-then-delete for classify, extract, lookup, and standardize models, plus a test that a non-OK delete response raises `RuntimeError`.

### DIFF
--- a/tests/connectors/test_train.py
+++ b/tests/connectors/test_train.py
@@ -1,7 +1,5 @@
 import uuid
 
-from pytest_mock import mocker
-
 import wrangles
 import pandas as pd
 import pytest
@@ -16,6 +14,21 @@ class LogCapture(logging.Handler):
         self.records.append(record)
     def get_messages(self):
         return [self.format(r) for r in self.records]
+
+def _delete_model_from_log(caplog, log_marker):
+    """
+    Find a model id logged as "<log_marker> :: <id>" and delete it, so tests
+    that create a real model clean up after themselves. Tolerates failure
+    while the real delete endpoint is unavailable.
+    """
+    for record in caplog.records:
+        if record.levelname == "INFO" and log_marker in record.message:
+            model_id = record.message.split("::")[-1].strip()
+            try:
+                wrangles.train.delete(model_id)
+            except RuntimeError:
+                pass
+            return
 
 #
 # Classify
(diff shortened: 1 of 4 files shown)


ebhills:
@mborodii-prog 
 - [ ] Add ability to call delete from recipe. Add ```confirm: delete``` parameter that requires writer to type delete
 - [ ] How is this different from PR #997 ?

ebhills (changes requested):
- [ ] Add ability to call delete from recipe. Add confirm: delete parameter that requires writer to type delete
- [ ] How is this different from PR https://github.com/wrangleworks/WranglesPY/pull/997 ?

ebhills:
Queue triage (2026-07-27)

- **Next action:** Rescope this around test-model cleanup. The public `train.delete` API overlaps #997 and this branch contains unrelated changes; build the cleanup as a focused follow-up after #997 or close this PR.

ebhills:
@copilot resolve the merge conflicts in this pull request

ebhills (approved):

PR #1070, opened 2026-07-17
942 new variable user permission team

### LINKED ISSUE #942: new Variable: user_permission_team
We need a way to set recipe variables based on the user running it.  But we need ability to do this based on the ORG or Team they belong to that has granted them access.  

Let's add a new system variable that can be used by Recipes.

Image

proposing we call this the user_permission_team, but open to a better name (@thomasstvr ).

 - the value should be the string name (not some id)
 - it should follow our hierarchy logic so that the ORG / TEAM that gives them the highest access is returned

mborodii-prog:
## Summary

Adds a recipe variable named `user_permission_team`.

This variable is intended to expose the permission-driving team for the current recipe run, so recipes can branch behavior based on the user's effective permission team.

## Problem

Some recipes need to apply different limits depending on the user running them.

For example, a web research recipe may allow one group to process more rows than another group. The row limits themselves should remain recipe logic, but the recipe needs access to the current user's effective permission group.

## Changes

- Added `user_permission_team` as an automatic recipe variable.
- For remote recipe runs, `user_permission_team` is read from recipe metadata when the backend provides it.
- For local/direct runs, `user_permission_team` falls back to the authenticated token claim named `user_permission_team`.
- Explicit caller-provided variables still take precedence.
- Added tests for templated values, `if` conditions, explicit overrides, and remote recipe metadata.

## Backend Contract

WranglesPY expects the effective permission group to be provided as:

```text
user_permission_team
```

The same name is exposed inside recipes.

## Usage Examples

Use in an `if` condition:

```yaml
wrangles:
  - create.column:
      output: allowed
      value: true
      if: user_permission_team == 'admin'
```

Use as a templated value:

```yaml
read:
  - test:
      rows: 1
      values:
        group: ${user_permission_team}
```

Use to choose a row limit in recipe variables:

```yaml
variables:
  max_rows: 100

wrangles:
  - filter:
      if: user_permission_team == 'small_team'
      where: index < ${max_rows}
```

### DIFF
--- a/tests/connectors/test_train.py
+++ b/tests/connectors/test_train.py
@@ -1400,11 +1400,25 @@ class TestTrainLookup:
         assert 'Blade Runner Upsert' in df['City'].values
         assert 'New Value' in df['City'].values
     
-    def test_missing_columns_error_message(self):  
+    def test_missing_columns_error_message(self, monkeypatch):  
         """  
         Verify that INSERT/UPSERT/UPDATE raise the expected error  
         when incoming columns are not present in the existing model.  
         """  
+        train_connector = importlib.import_module("wrangles.connectors.train")
+        monkeypatch.setattr(
+            train_connector._data,
+            "model",
+            lambda model_id: {"variant": "key"}
+        )
+        monkeypatch.setattr(
+            train_connector._data,
+            "model_content",
+            lambda model_id: {
+                "Columns": ["Key", "Value"],
+                "Data": [["k1", "v1"], ["k2", "v2"]],
+            }
+        )
 
         df = pd.DataFrame({  
             "Key": ["k3"],  
--- a/tests/connectors/test_train.py
+++ b/tests/connectors/test_train.py
@@ -1,12 +1,37 @@
 import uuid
 
-from pytest_mock import mocker
-
 import wrangles
 import pandas as pd
 import pytest
 import logging
 import re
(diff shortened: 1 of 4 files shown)


ebhills:
Queue triage (2026-07-27)

- **Next action:** Validate and document the backend/token contract for `user_permission_team`, keep explicit-override coverage, then request one primary reviewer.

ebhills:
@copilot+mai-code-1.1-flash:medium - 2 things:
 - in a previous review the agent changed the name of tests and functions from user_group to user_permission_team, probably because of the title of this PR.  However, we had decided to refer to this new variable as user_group because 'group' refers to both teams (childern) and orgs (parents), either of which can determine authorization. Upon further consideration, let's change all references to applied_permission_group.  this makes its meaning clear.
 - resolve the conflicts.

ebhills:
@mborodii-prog - please see my recent comments about the name for this variable.  I have decided to call it:
`applied_permission_group`

Please review this here and in `API-Core` & `wranglesXL` to ensure we are consistent across all 3. And let's make sure it's visible in XL (via inspect)

mborodii-prog:
@ebhills I updated current PR, also now I'm working on XL side changes. Do we have some default value for that new variable?

ebhills (approved):

PR #1093, opened 2026-07-27
Fix recipe schema so nested recipe: wrangles accept where

mborodii-prog:
## Summary
- The `recipe` wrangle (used to run a sub-recipe/"recipe-ception") is the only wrangle whose docstring schema uses a top-level `anyOf` instead of a plain `properties` block, since it's recursive.
- `schema/generate_recipe_schema.py`'s common-property injection loop assumed every wrangle has a top-level `properties` key, so `where`/`where_params`/`if` were silently added to an orphan `properties` dict that the validator never consulted for `recipe`.
- Result: valid recipes using `where` on a nested `recipe:` wrangle were rejected by the generated JSON schema (IDE/editor validation), even though it works correctly at runtime.

## Fix
- `schema/generate_recipe_schema.py`: fall back to merging common properties into the last `anyOf` branch's `properties`, matching the existing pattern already used for `read`/`write` wrangles.
- `schema/recipe_base_schema.json`: added `where`/`where_params`/`if` as optional root properties, since the recursive `$ref: "#"` branch (inline nested recipe) points back at the root schema.

### DIFF
--- a/schema/generate_recipe_schema.py
+++ b/schema/generate_recipe_schema.py
@@ -158,24 +158,28 @@ getMethodDocs(schema['wrangles'], wrangles.recipe._recipe_wrangles, '')
 
 # Add common wrangle properties
 for wrangle in schema['wrangles']:
-    if "properties" not in schema['wrangles'][wrangle]:
-        schema['wrangles'][wrangle]["properties"] = {}
+    if "properties" in schema['wrangles'][wrangle]:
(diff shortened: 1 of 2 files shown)


ebhills:
Queue triage (2026-07-27)

- **Next action:** Held in Draft only to enforce the five-PR review WIP limit. When a Ready slot opens, update from `main`, confirm schema-generation checks, and request one primary reviewer.

ebhills (approved):

